PR #4, opened 2021-05-12
Create ceres-early-investors-sale pallet

MasterRadule:
Pallet for Ceres token early investors sale


takemiyamakoto:
Please write unit tests

PR #258, opened 2023-02-08
Fix TBCD and add tests

vovac12:


pallets/multicollateral-bonding-curve-pool/src/lib.rs
```diff
@@ -1446,7 +1459,7 @@ impl<T: Config> Pallet<T> {
         let current_state =
             Self::buy_function(&base_asset_id, collateral_asset_id, price_variant, delta)?;
 
-        let price = (initial_state + current_state) / fixed_wrapper!(2.0)
+        let price = ((initial_state + current_state) / fixed_wrapper!(2.0))
```
wer1st, line 1462: Looks like redundant

wer1st (approved):

pakrentos (approved):

PR #277, opened 2023-02-16
Merge master to bridge

wer1st:


pallets/hermes-governance-platform/src/benchmarking.rs
```diff
@@ -0,0 +1,246 @@
+//! Hermes governance platform module benchmarking.
```
pakrentos, line 1: Not sure, if it is important during this merge, but there is missing copyright in hermes-governance-platform pallet files

pallets/demeter-farming-platform/src/tests.rs
```diff
@@ -2394,17 +2394,6 @@ mod tests {
                     deposit_fee,
                 )
             );
-
```
vovac12, line 2397: Seems like this code should not be removed

vovac12 (changes requested):

pallets/hermes-governance-platform/src/benchmarking.rs
wer1st: As I remember, hermes is not our code and I'm not sure we can add our copyright header for these files. The same logic for ceres.

vovac12 (approved):

pakrentos (approved):

Issue #255, opened 2023-02-07
[Feature]: Show route to the liquidityProxy.quote RPC


Tieumsan:
### As a

Developer / User

### I want to

Know the asset routing for my swap quote

### So that

I can easily display the swap routing in the UI

### Preconditions

User has inputted the parameters `inputAssetId`, `outputAssetId`, `amount`.

The user/system has sent the RPC call `liquidityProxy.quote`

### Flow

1. The user has entered the parameters `inputAssetId`, `outputAssetId`, `amount`
2. The external system (client application) send the RPC call `liquidityProxy.quote(inputAssetId, outputAssetId, amount)`
3. The system (blockchain) returns the swap routing.
  - object name: `route`
  - type: [string]
  - string format: AssetId in form of "0x..."

ra9mls:
![image.png](https://images.zenhubusercontent.com/62bbde011ca04a42bf2c5247/104c381b-77e1-44df-8cb0-51ff324211b5)

ra9mls:
Nodes need to be updated

PR #283, opened 2023-02-21
Quote rpc route

bragov4ik:
#255
Return exchange route on `quote` rpc


pallets/liquidity-proxy/src/lib.rs
```diff
@@ -709,7 +691,12 @@ impl<T: Config> Pallet<T> {
                     deduce_fee,
                 ),
             };
-            quote.map(|x| (x, atomic_path))
+            quote.map(|x| QuoteInfo {
```
pakrentos, line 694: May be a good idea to add a short function for QuoteInfo which will build itself based upon QuoteOutcome

bragov4ik: Not sure what is QuoteOutcome.

If you mean building from tuple as done in this line, then I don't think there is a need to make a separate function for this. such conversion (and any construction of struct `QuoteInfo`) occurs only here and `path` is added apart from the tuple

pakrentos: @bragov4ik updated prev. comment; I mean add function like this
```
impl<T: Config> QuoteInfo<T> {
    pub fn from_parts(amount: QuoteAmount<T>, path: Vec<T::AssetId>) -> Self {
        Self {
            outcome: amount.0,
            rewards: amount.1,
            liquidity_sources: amount.2,
            path: path,
         }
    }
}
```
Just a suggestion related to code clarity.

bragov4ik: Do you suggest to create a new struct for return value of `quote_pairs_with_flexible_amount`?

sevenzing (approved):

vovac12 (approved):

PR #306, opened 2023-03-06
Remove network data extrinsic

vovac12:
Add extrinsic to remove eth-bridge network. Mostly needed to remove Rinkeby network from stage and replace it with Sepolia


vovac12:
This problem is solved in another way, so we don't need such extrinsic now

PR #320, opened 2023-03-08
Fix/swap transfer batch slippage hotfix

pakrentos:
# Changes
- Fixed slippage check
- Added temporary solution for inner_exchange inaccurate exchange result

## Notes
inner_exchange executed output amount does not match desired output amount. It causes problems, if a caller does not hold any tokens requested in batches. To resolve this, I calculate a remainder (it's about 10^-15 tokens on average) and spread it between receivers.
I've used assets.totalBalance for remainder calculation, since inner_exchange's returned SwapOutcome holds info about executed input amount, not the output amount, as it stated in docs (see liquidityProxy.exchangeSequence code).

The problem originates from the inaccurate quote results, which are used for calculating input amount in WithDesiredOutput swap variant.

### DIFF
--- a/pallets/liquidity-proxy/src/lib.rs
+++ b/pallets/liquidity-proxy/src/lib.rs
@@ -1784,51 +1784,66 @@ pub mod pallet {
                         receivers,
                     } = swap_batch_info;
 
-                    let filter = LiquiditySourceFilter::with_mode(
-                        dex_id,
-                        filter_mode.clone(),
-                        selected_source_types.clone(),
-                    );
-
-                    if Self::is_forbidden_filter(
-                        &input_asset_id,
-                        &asset_id,
-                        &selected_source_types,
-                        &filter_mode,
-                    ) {
-                        fail!(Error::<T>::ForbiddenFilter);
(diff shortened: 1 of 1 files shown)


wer1st (dismissed):

pallets/liquidity-proxy/src/lib.rs
```diff
@@ -1784,53 +1784,79 @@ pub mod pallet {
...
+                                filter.clone(),
+                            )?;
+
+                            let caller_output_asset_balance =
+                                assets::Pallet::<T>::total_balance(&asset_id, &who)?;
+                            let remainder_per_receiver: Balance =
+                                if caller_output_asset_balance < out_amount {
+                                    let remainder =
+                                        out_amount.saturating_sub(caller_output_asset_balance);
+                                    remainder / (receivers.len() as u128)
```
vovac12, line 1834: Add check for empty `receivers`

vovac12 (approved):

vovac12 (approved):

wer1st (approved):

PR #327, opened 2023-03-11
Liquidity Proxy runtime weights

wer1st:
Now liquidity-proxy calculates the weight dynamically.


pallets/liquidity-proxy/src/lib.rs
```diff
@@ -1053,6 +1122,164 @@ impl<T: Config> Pallet<T> {
...
+    ///                 exchange() - call N times, where N is a count of assets in the path
+    ///
+    /// Dev NOTE: if you change the logic of liquidity proxy, please sustain inner_exchange_weight() and code map above.
+    pub fn inner_exchange_weight(
+        dex_id: &T::DEXId,
+        input: &T::AssetId,
+        output: &T::AssetId,
+        swap_variant: SwapVariant,
+    ) -> Weight {
+        let dex_info = dex_manager::Pallet::<T>::get_dex_info(dex_id).unwrap();
```
N1ghtStorm, line 1174: Are there any cases where these unwrap calls may cause panic?

pallets/liquidity-proxy/src/lib.rs
```diff
@@ -1053,6 +1122,164 @@ impl<T: Config> Pallet<T> {
...
+        let tp = ExchangePath::<T>::new_trivial(&dex_info, *input, *output).unwrap();
```
N1ghtStorm, line 1175: for example here  https://github.com/sora-xor/sora2-network/blob/b10212041c718bf8ea842c6aa5fa9ae208e046da/pallets/liquidity-proxy/src/lib.rs#L1291 path can be None and error scenarios are written

N1ghtStorm (changes requested):

pallets/liquidity-proxy/src/lib.rs
wer1st: The same ways when it can panic inside the code.
It seems that in this case there is no difference when the panic will be triggered, sooner or later.

pallets/liquidity-proxy/src/lib.rs
wer1st: If `new_trivial ` returns `None` we want to get panic, I guess

N1ghtStorm (approved):

vovac12 (approved):

PR #338, opened 2023-03-13
Fix fee shows in output asset instead of XOR

bragov4ik:


pallets/xst/src/lib.rs
```diff
@@ -704,6 +725,16 @@ impl<T: Config> LiquiditySource<T::DEXId, T::AccountId, T::AssetId, Balance, Dis
...
     }
 
+    /// ## Implementation
+    ///
+    /// The implementation of this method might be quite confusing,
+    /// [here](https://hackmd.io/@jTXlKyDTTYuWUChtsJYcCg/SJ5ItHw12)
+    /// you can find the main logic and formulae without too
+    /// much technical details
+    ///
+    /// Unfortunately, confluence and rustdoc do not render math
```
vovac12, line 735: I think we can try to add formulas support to our project. It also would be useful for another pallets docs
https://docs.rs/rustdoc-katex-demo/0.1.5/rustdoc_katex_demo/

pallets/xst/src/lib.rs
```diff
@@ -721,6 +752,26 @@ impl<T: Config> LiquiditySource<T::DEXId, T::AccountId, T::AssetId, Balance, Dis
         } else {
             Self::decide_buy_amounts(&output_asset_id, &input_asset_id, amount, deduce_fee)?
         };
+        let fee_amount = if deduce_fee {
+            // `fee_amount` is always computed to be in `main_asset_id`, which is
+            // `SyntheticBaseAssetId` (e.g. XST), but `SwapOutcome` assumes XOR
+            // (`BaseAssetId`), so we convert.
+            let output_to_base: FixedWrapper =
+                <T as pallet::Config>::PriceToolsPallet::get_average_price(
```
vovac12, line 760: I think we should add check for `input_asset == output_asset` to `get_average_price` function

pallets/xst/src/lib.rs
bragov4ik: image
Added & checked locally

pakrentos (approved):

vovac12 (approved):

vovac12 (approved):

PR #351, opened 2023-03-21
Update v0.9.38

wer1st:


common/src/weights.rs
```diff
@@ -58,7 +58,9 @@ pub struct PresetWeightInfo<T>(PhantomData<T>);
 /// by  Operational  extrinsics.
 const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
 /// We allow for 2 seconds of compute with a 6 second average block time.
-const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND.saturating_mul(2).set_proof_size(2);
+const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_ref_time(WEIGHT_REF_TIME_PER_SECOND)
```
N1ghtStorm, line 61: I propose to check this parameter. I had a problem on the parachain with the same parameter - transactions could not execute due to block size exceeded. Tried to run the mainnet from this branch and faced the same problem. Please check it. 
I solved the same issue using: https://github.com/sora-xor/sora2-parachain/blob/55575694a6f622f33e1228738c5a0be279433427/runtime/src/lib.rs#L266

but it seems like it could not be suitable in this case

housekeeping/docker/env/Dockerfile
```diff
@@ -44,7 +44,7 @@ RUN add-apt-repository \
     rm -rf /var/lib/apt/lists/*
 
 # Install rust
-ENV RUST_NIGHTLY_VERSION=nightly-2022-05-12
+ENV RUST_NIGHTLY_VERSION=nightly-2022-10-25
```
f33r0, line 47: 2023-03-21

common/src/weights.rs
wer1st: It is PR about updating Substrate, and I just updated the const.
Your suggestion is good to investigate, but is out of scope of this task.
Please create a separate ticket for that.

wer1st: Updated to value from template

vovac12 (approved):

PR #368, opened 2023-03-27
Order book pallet layout

bragov4ik:


wer1st:
Will be changed and open later

PR #388, opened 2023-04-06
Order book skeleton

wer1st:


pallets/order-book/Cargo.toml
```diff
@@ -0,0 +1,62 @@
...
+	"pallet-timestamp/std",
+	"scale-info/std",
+	"sp-runtime/std",
+	"sp-std/std",
+	"technical/std",
+	"trading-pair/std",
+]
+
+runtime-benchmarks = [
+	"frame-benchmarking",
```
bragov4ik, line 52: Spaces & tabs got mixed up somehow

pallets/order-book/src/order_book.rs
```diff
@@ -0,0 +1,58 @@
...
+pub enum OrderBookStatus {
+    Trade,
+    PlaceAndCancel,
+    OnlyCancel,
+    Stop,
+}
+
+#[derive(Encode, Decode, Clone, Debug, scale_info::TypeInfo, MaxEncodedLen)]
+#[scale_info(skip_type_params(T))]
+pub struct OrderBook<T>
```
bragov4ik, line 46: Maybe rename to `OrderBookSettings` or smth since we don't actually store orders inside? To avoid confusion

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,439 @@
...
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
+        #[pallet::weight(<T as Config>::WeightInfo::create_orderbook())]
+        pub fn create_orderbook(
+            origin: OriginFor<T>,
+            _dex_id: T::DEXId,
+            order_book_id: OrderBookId<T>,
+            _tick_size: T::Balance,
```
bragov4ik, line 229: It seems that this type is not used anywhere else in this repo. Should it be just `Balance` instead?

pallets/order-book/src/limit_order.rs
```diff
@@ -0,0 +1,50 @@
...
+where
+    T: crate::Config,
+{
+    pub id: T::OrderId,
+    pub owner: T::AccountId,
+    pub side: PriceVariant,
+    pub price: T::Balance,
+    pub original_amount: T::Balance,
+    pub executed_amount: T::Balance,
+    pub time: T::Moment,
```
bragov4ik, line 48: Maybe rename so that it's more clear what `time` it is (creation time?)

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+        /// Price in given order exceeds allowed limits for the trading pair
+        PriceExceedsLimits,
+        /// Lifespan exceeds defined limits
+        InvalidLifespan,
+        /// Order book does not exist for this trading pair
+        UnknownOrderBook,
+        /// Order book already exists for this trading pair
+        OrderBookAlreadyExists,
+        /// Cannot insert the limit order
+        InsertLimitOrderError,
```
bragov4ik, line 214: Apparently, this error is only returned when the limit of bounded vector is reached. I think it will be better if this is represented in the error somehow. Otherwise this (and `DeleteLimitOrderError`) seems pretty uninformative

pallets/order-book/src/tests.rs
```diff
@@ -0,0 +1,326 @@
...
+type OrderBook = Pallet<Runtime>;
+
+fn alice() -> <Runtime as frame_system::Config>::AccountId {
+    <Runtime as frame_system::Config>::AccountId::new([1u8; 32])
+}
+
+type E = order_book::Error<Runtime>;
+
+#[test]
+fn insert_limit_order_success() {
```
bragov4ik, line 48: Might be a nitpicking, but I think it's better to name tests uniformly across the pallets. For example, we have names like `should_...` or `should_not_..._if_...` in other pallets. Maybe follow the same strategy, so the code looks consistent?

pallets/order-book/src/limit_order.rs
wer1st: Yes, and it is standart name of this field for orders in different exchanges

pallets/order-book/src/lib.rs
wer1st: Types will be changed with my next commit

wer1st: And no, we should use T::Balance instead of Balance in all pallets

pallets/order-book/Cargo.toml
wer1st: Yes, autoformatter don't fix toml

pallets/order-book/src/order_book.rs
wer1st: No, it will contain order book logic inside. Leave as is

pallets/order-book/src/lib.rs
wer1st: Will fix Insert Error, but leave DeleteLimitOrderError as is. This error should never happen because we will check that order exists before deleting

pallets/order-book/src/tests.rs
wer1st: Ok

pallets/order-book/src/lib.rs
bragov4ik: Then [this type](https://github.com/sora-xor/sora2-network/pull/388/files#diff-2caeaf0a4f9f693d5b36b62bcf857021fa00942c659d2a90f126238bf6c99270R290) should be changed to T::Balance (just in case it's missed)

pallets/order-book/src/lib.rs
bragov4ik: > we will check that order exists before deleting
So all 12 occurrences of this error are impossible to get returned? I'm just worried that if we miss something it will be pretty hard to debug

wer1st: I think the list of 12 different errors for this case is redundant. It should never happen, if it is it means we have big problem in our code related to data consistency.
Moreover, these errors are returned to the user. They must be informative from the user perspective first of all, secondly - for debug.

wer1st: Added comments

bragov4ik (approved):
LGTM

runtime/src/lib.rs
```diff
@@ -1949,6 +1949,18 @@ impl hermes_governance_platform::Config for Runtime {
     type WeightInfo = hermes_governance_platform::weights::WeightInfo<Runtime>;
 }
 
+#[cfg(feature = "wip")] // order-book
+impl order_book::Config for Runtime {
+    const MAX_ORDER_LIFETIME: Moment = 2_592_000_000; // 30 days
```
vovac12, line 1954: I think it's better to use predefined constants for it

runtime/src/lib.rs
```diff
@@ -1949,6 +1949,18 @@ impl hermes_governance_platform::Config for Runtime {
...
+    const MAX_OPENED_LIMIT_ORDERS_COUNT: u32 = 100;
+    type RuntimeEvent = RuntimeEvent;
+    type OrderId = u128;
+    type MaxOpenedLimitOrdersForAllOrderBooksPerUser = ConstU32<10000>;
```
vovac12, line 1958: We will need to benchmark almost full order books in future

relayer/substrate-gen/Cargo.toml
```diff
@@ -12,8 +12,6 @@ scale-info = { version = "2", default-features = false, features = ["derive"] }
 framenode-runtime = { path = "../../runtime", default-features = false, features = [
     "std",
     "private-net",
-    "wip",
```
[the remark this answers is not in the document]
wer1st: relayer looks like component that should work with "release" version of framenode-runtime

runtime/src/lib.rs
wer1st: We have predefined constants for blocks per day. We have no something to have secs per day

pallets/order-book/src/lib.rs
```diff
@@ -0,0 +1,544 @@
...
+    ) -> Result<(), DispatchError> {
+        <LimitOrders<T>>::insert(order_book_id, order.id, order);
+
+        match order.side {
+            PriceVariant::Buy => {
+                <Bids<T>>::try_append(order_book_id, order.price, order.id)
+                    .map_err(|_| Error::<T>::LimitOrderStorageOverflow)?;
+
+                let mut bids = <AggregatedBids<T>>::get(order_book_id);
+                let volume = bids.get(&order.price).map(|x| *x).unwrap_or_default() + order.amount;
```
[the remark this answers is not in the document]
wer1st: Fixed in #396 becuase Balance type is changed and it is easier to change in #396

relayer/substrate-gen/Cargo.toml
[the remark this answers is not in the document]
vovac12: It's mostly used for trustless bridges and they will be hidden with `wip` feature in #398

runtime/src/lib.rs
vovac12: You can use `30 * (DAYS as Moment) * SECS_PER_BLOCK`. I think it looks better than much magic number

vovac12 (approved):

PR #398, opened 2023-04-12
Add feature flags for bridges

vovac12:


node/chain_spec/src/lib.rs
```diff
@@ -1143,18 +1147,28 @@ fn testnet_genesis(
     ];
     let initial_synthetic_assets = vec![XSTUSD.into()];
     GenesisConfig {
+        #[cfg(feature = "wip")]
```
wer1st, line 1150:
```suggestion
        #[cfg(feature = "wip")] // bridge
```

wer1st: Or other name.
Please use the comment for all your flags.
It makes more easy to recognize the feature scope of every single flag

vovac12: Added everywhere except imports

wer1st (approved):

Issue #318, opened 2023-03-06
Locked tokens


bgodlin:
The user's free tokens will be transferred to the respective orderbook balance when placing orders, and will only become available again when they are withdrawn from the orderbook.

bragov4ik:
need to have tech acc for each pair

bragov4ik:
Status was reset after issue with github, not sure if already tested

PR #401, opened 2023-04-13
Order book lock tokens

bragov4ik:
#318


node/chain_spec/src/lib.rs
```diff
@@ -1811,6 +1824,17 @@ fn mainnet_genesis(
         technical::Pallet::<Runtime>::tech_account_id_to_account_id(&dex_root_tech_account_id)
             .unwrap();
 
+    #[cfg(feature = "wip")]
```
bragov4ik, line 1827: I'm not sure if it should be done this way. Should I remove the cfg at all?? Or we need to exclude WIP stuff from main chainspec

pallets/eth-bridge/src/tests/mock.rs
```diff
@@ -132,6 +132,13 @@ impl<RuntimeCall: Codec + Sync + Send, Context, Extra> Checkable<Context>
     fn check(self, _c: &Context) -> Result<Self::Checked, TransactionValidityError> {
         Ok(self)
     }
+
+    fn unchecked_into_checked_i_know_what_i_am_doing(
```
bragov4ik, line 136: it complained because of `try-runtime` flag

pallets/order-book/src/tests/pallet.rs
```diff
@@ -450,3 +454,224 @@ fn should_not_delete_limit_order() {
...
+        ));
+        let balance_before =
+            assets::Pallet::<Runtime>::free_balance(&XOR, &alice()).expect("XOR must exist");
+
+        assert_ok!(OrderBook::lock_liquidity(&alice(), &XOR, amount_to_lock));
+
+        let balance_after_lock =
+            assets::Pallet::<Runtime>::free_balance(&XOR, &alice()).expect("XOR must exist");
+        let locked = balance_before - balance_after_lock;
+        assert!(amount_to_lock == locked);
```
wer1st, line 477:
```suggestion
        assert_eq!(balance_after_lock, balance_before - amount_to_lock);
```

pallets/order-book/src/tests/pallet.rs
```diff
@@ -450,3 +454,224 @@ fn should_not_delete_limit_order() {
...
+
+        assert_ok!(OrderBook::unlock_liquidity(&alice(), &XOR, amount_to_lock));
+
+        let balance_after_unlock =
+            assets::Pallet::<Runtime>::free_balance(&XOR, &alice()).expect("XOR must exist");
+        assert!(balance_before == balance_after_unlock);
```
wer1st, line 483:
```suggestion
        assert_eq!(balance_before, balance_after_unlock);
```

pallets/order-book/src/tests/pallet.rs
```diff
@@ -450,3 +454,224 @@ fn should_not_delete_limit_order() {
...
+        ));
+        let balance_before =
+            assets::Pallet::<Runtime>::free_balance(&VAL, &alice()).expect("VAL must exist");
+
+        assert_ok!(OrderBook::lock_liquidity(&alice(), &VAL, amount_to_lock));
+
+        let balance_after_lock =
+            assets::Pallet::<Runtime>::free_balance(&VAL, &alice()).expect("VAL must exist");
+        let locked = balance_before - balance_after_lock;
+        assert!(amount_to_lock == locked);
```
wer1st, line 506: The same as above

pallets/order-book/src/tests/pallet.rs
```diff
@@ -450,3 +454,224 @@ fn should_not_delete_limit_order() {
...
+            None,
+            None,
+        )
+        .unwrap();
+
+        assert_ok!(OrderBook::lock_liquidity(&alice(), &nft, balance!(1)));
+
+        let balance_after_lock =
+            assets::Pallet::<Runtime>::free_balance(&nft, &alice()).expect("NFT must exist");
+        assert!(balance_after_lock == balance!(0));
```
wer1st, line 537:
```suggestion
        assert!(balance_after_lock.is_zero());
```

pallets/order-book/src/tests/pallet.rs
wer1st, line 1: Test the case when asset is locked for Alice and then unlocked for Bob.
It is a mechanism of exchange: Alice locks the asset in the limit order, then Bob buys/sells it and takes the locked liquidity.

runtime/src/lib.rs
```diff
@@ -1965,6 +1972,7 @@ impl order_book::Config for Runtime {
     type EnsureTradingPairExists = TradingPair;
     type AssetInfoProvider = Assets;
     type DexInfoProvider = DEXManager;
+    type LockTechAccountId = GetOrderBookLockTechAccountId;
```
wer1st, line 1975: @vovac12 What do you think, is it ok to have 1 tech account for all order books.
Mathematically it's ok, but we have the mixed liquidity of all order books in 1 account.

node/chain_spec/src/lib.rs
wer1st: I guess we should to hide the code above under the `wip` feature flag as well.

runtime/src/lib.rs
bragov4ik: It seems that having one for each trading pair or something similar is not possible, because of the complicated process of registering the accounts (and having to do runtime migrations for it). However yes, let's wait for Vladimirs answer, maybe I'm wrong

pallets/order-book/Cargo.toml
```diff
@@ -33,6 +33,7 @@ sp-core = { git = "https://github.com/sora-xor/substrate.git", branch = "polkado
 sp-io = { git = "https://github.com/sora-xor/substrate.git", branch = "polkadot-v0.9.38", default-features = false }
 framenode-chain-spec = { path = "../../node/chain_spec", features = ["test"] }
 framenode-runtime = { path = "../../runtime", default-features = false, features = ["std"] }
+pallet-balances = { git = "https://github.com/sora-xor/substrate.git", branch = "polkadot-v0.9.38" }
```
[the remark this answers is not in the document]
bragov4ik: It's actually needed to assert on errors from the pallet

pallets/order-book/src/benchmarking.rs
```diff
@@ -33,6 +33,7 @@
 #![cfg(feature = "runtime-benchmarks")]
 // order-book
 #![cfg(feature = "wip")]
+// order-book
```
wer1st, line 36: there is a comment above 😃 
auto-formatter doesn't allow to have 1 line here.
I decided to move it before feature flag to not confuse anyone that the comment is related to `#![cfg(not(test))]` below.
btw it is minor, just decide and leave only one comment above or below

wer1st (approved):

runtime/src/lib.rs
vovac12: You're right, it's better to have tech account for each trading pair. If we will register tech accounts during order book registration we will not need migrations at all. I think we should add new variant to `TechPurpose` enum and create order book tech account with it

pallets/order-book/src/benchmarking.rs
bragov4ik: I've added it on the same line to follow development process. Fmt moved it there :) Fixed

runtime/src/lib.rs
bragov4ik: I suppose they also should be deregistered on removal of trading pair, right?

pallets/order-book/src/lib.rs
```diff
@@ -553,23 +563,82 @@ impl<T: Config> Pallet<T> {
 
         Ok(())
     }
+}
+
+// todo: make pub(tests) (k.ivanov)
+pub trait CurrencyLocker<AccountId, AssetId, DEXId> {
```
bragov4ik, line 569: tbh I don't think trait is needed here; it doesn't contribute to loose coupling, just adds redundant code

bragov4ik: @wer1st what do you think?

wer1st: You can leave it here, I will move it in `traits.rs` in my branch

pallets/order-book/src/lib.rs
```diff
@@ -553,23 +563,87 @@ impl<T: Config> Pallet<T> {
...
+    /// Lock `amount` of liquidity in `trading_pair`'s asset chosen by `asset`
+    fn lock_liquidity(
+        dex_id: DEXId,
+        account: &AccountId,
+        trading_pair: common::TradingPair<AssetId>,
+        asset: common::TradingPairSelector,
+        amount: Balance,
+    ) -> Result<(), DispatchError>;
+
+    fn unlock_liquidity(
```
wer1st, line 579: Add description as above

pallets/order-book/src/lib.rs
```diff
@@ -553,23 +563,87 @@ impl<T: Config> Pallet<T> {
...
+    // todo: make pub(tests) (k.ivanov)
+    /// Validity of asset ids (for example, to have the same base asset
+    /// for dex and pair) should be done beforehand
+    pub fn deregister_tech_account(
+        dex_id: T::DEXId,
+        trading_pair: OrderBookId<T>,
+    ) -> Result<(), DispatchError> {
+        let tech_account = Self::tech_account_for_order_book(dex_id, trading_pair);
+        technical::Pallet::<T>::register_tech_account_id(tech_account)
     }
```
wer1st, line 647: The same functions

pallets/order-book/src/lib.rs
bragov4ik: Ok, my concern was about whether it is needed at all. However, now I think about it, it will be provided to order book datastructure, so it makes sense at the end

pallets/order-book/src/lib.rs
bragov4ik: Oops, thanks!

common/src/primitives.rs
```diff
@@ -116,6 +115,35 @@ impl<AssetId: Eq> TradingPair<AssetId> {
...
+    }
+}
+
+/// Enum to help choosing asset from [`TradingPair`].
+///
+/// Should help when we need to use asset id that must be from the
+/// pair. Instead of passing the pair with id and checking
+/// `pair.consists_of(id)`, we can pass the pair and `TradingPairSelector`
+#[derive(Clone)]
+pub enum TradingPairSelector {
```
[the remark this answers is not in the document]
bragov4ik: This was the best approach I could come up with for situation described in the comments above. I'm open to other ideas

pallets/technical/src/lib.rs
```diff
@@ -157,25 +156,35 @@ impl<T: Config> Pallet<T> {
...
+        if let Err(_) = Self::lookup_tech_account_id(&account_id) {
+            frame_system::Pallet::<T>::inc_providers(&account_id);
+            TechAccounts::<T>::insert(account_id, tech_account_id.clone());
+        }
+        Ok(())
+    }
+
+    // todo: make pub(tests) (k.ivanov)
+    /// Deregister `TechAccountId` in storage map.
+    pub fn deregister_tech_account_id(tech_account_id: T::TechAccountId) -> DispatchResult {
```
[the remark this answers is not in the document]
bragov4ik: They're pretty much interchangeable. However, 'de-' is more often used to describe active action of removal (from register). Therefore, `deregister` seems more appropriate IMO.

wer1st (approved):

pallets/technical/src/lib.rs
```diff
@@ -275,9 +284,12 @@ pub mod pallet {
             + FromGenericPair
             + MaybeSerializeDeserialize
             + common::ToFeeAccount
-            + common::ToTechUnitFromDEXAndTradingPair<
+            + common::ToXykTechUnitFromDEXAndTradingPair<
```
vovac12, line 287: Do we really need this trait bounds here?

bragov4ik: Discussed in DMs, haven't found a good way to move the bounds to dependant pallets

vovac12 (approved):

bragov4ik:
Status was reset after issue with github, not sure if already tested

PR #403, opened 2023-04-14
Fix runtime Config

vovac12:


pakrentos (approved):

runtime/src/lib.rs
```diff
@@ -316,7 +316,8 @@ parameter_types! {
     pub const TechnicalCollectiveMotionDuration: BlockNumber = 5 * DAYS;
     pub const TechnicalCollectiveMaxProposals: u32 = 100;
     pub const TechnicalCollectiveMaxMembers: u32 = 100;
-    pub const SchedulerMaxWeight: Weight = Weight::from_parts(1024, 0);
+    pub const SchedulerMaxWeight: Weight = Perbill::from_percent(60) * BlockWeights::get().max_block;
```
wer1st, line 319:
```suggestion
    pub const SchedulerMaxWeight: Weight = BlockWeights::get().max_block.saturating_mul(Perbill::from_percent(60));
```

vovac12: This method is not implemented for this types. And mul implementation for Perbill is working without overflows

wer1st (approved):

PR #417, opened 2023-04-20
Mark bridges as ready to test

vovac12:


node/src/rpc.rs
```diff
@@ -119,6 +112,13 @@ where
     Ok(rpc)
 }
 
+#[cfg(feature = "wip")]
+pub fn add_wip_rpc(
+    rpc: RpcExtension,
+) -> Result<RpcExtension, Box<dyn std::error::Error + Send + Sync>> {
+    Ok(rpc)
+}
+
```
wer1st, line 121: Why do you need it if your feature will work only under `ready-to-test`?

vovac12: I decided to keep it empty for future usage. For example if we will add RPC to order-book or to another wip feature. To not remove and add this function every time

wer1st: Ok

bragov4ik (approved):

PR #423, opened 2023-04-24
Fixes for xst and band pallets

pakrentos:
# Changes
- Fixed oracle symbol price calculation in XSTPool
- Refactored XSTPool docs
- Refactored XSTPool tests
- Fixed XOR fee calculation in XSTPool
- Added indivisibility check in set_reference_asset and in enable_synthetic_asset_unchecked for XSTPool
- Added asset existance check in set_reference_asset
- Changed units of time to seconds in band


vovac12 (commented):
Maybe we should use Duration or Moment everywhere to not depend on Band requests time precision

wer1st (approved):

vovac12 (approved):

PR #424, opened 2023-04-24
Fix CI

vovac12:


Jenkinsfile
```diff
@@ -149,8 +149,10 @@ pipeline {
                                     rm -rf ~/.cargo/.package-cache
                                     rm Cargo.lock
                                     cargo fmt -- --check > /dev/null
+                                    SKIP_WASM_BUILD=1 cargo check
+                                    SKIP_WASM_BUILD=1 cargo check --features private-net,ready-to-test
```
wer1st, line 153: Please remove this line. I think it is redundant, in this block we want to always check wip & ready-to-test and I suggest to omit this extra work.

vovac12: For test and stage environment we use only `ready-to-test` and we need to check it in PR CI instead of having failed CI for version tag. Also this checks is fast enough

f33r0 (approved):

wer1st (approved):

PR #444, opened 2023-05-14
Order Book: Compatibility with liquidityProxy

wer1st:


pallets/order-book/src/order_book.rs
```diff
@@ -213,6 +214,153 @@ impl<T: crate::Config + Sized> OrderBook<T> {
...
+            {
+                Ok(PriceVariant::Buy)
+            }
+            OrderBookId::<AssetIdOf<T>> { base, quote }
+                if base == *input_asset_id && quote == *output_asset_id =>
+            {
+                Ok(PriceVariant::Sell)
+            }
+            _ => Err(Error::<T>::InvalidAsset.into()),
+        }
```
bragov4ik, line 355:
```suggestion
        let OrderBookId::<AssetIdOf<T>> { base, quote } = self.order_book_id;
        if base == *output_asset_id && quote == *input_asset_id {
            Ok(PriceVariant::Buy)
        } else if base == *input_asset_id && quote == *output_asset_id {
            Ok(PriceVariant::Sell)
        } else {
            Err(Error::<T>::InvalidAsset.into()),
        }
```
should be easier to read

pallets/liquidity-proxy/src/lib.rs
```diff
@@ -250,17 +251,29 @@ impl<T: Config> Pallet<T> {
...
             && !selected_source_types
                 .contains(&LiquiditySourceType::MulticollateralBondingCurvePool)
             && !selected_source_types.contains(&LiquiditySourceType::XSTPool)
+            && !is_order_book
             && filter_mode == &FilterMode::AllowSelected
             || selected_source_types
                 .contains(&LiquiditySourceType::MulticollateralBondingCurvePool)
                 && selected_source_types.contains(&LiquiditySourceType::XSTPool)
                 && !selected_source_types.contains(&LiquiditySourceType::XYKPool)
+                && !is_order_book
```
vovac12, line 276:
```suggestion
                && is_order_book
```
I think it should be like this. But we definitely need to find better way to implement this condition

pallets/order-book/src/lib.rs
```diff
@@ -345,6 +358,16 @@ pub mod pallet {
                 OrderBook::<T>::default_nft(order_book_id, dex_id)
             };
 
+            #[cfg(feature = "wip")] // order-book
```
vovac12, line 361: This pallet is already hidden by `wip` feature, so you can pass `wip` feature to `common` crate in `Cargo.toml`

wer1st: The build is failed for `cargo build`, because `LiquiditySourceType::OrderBook` is hidden by `wip`

pallets/order-book/src/order_book.rs
wer1st: The current version looks readable as well

vovac12 (approved):

pallets/order-book/src/order_book.rs
bragov4ik: More nesting and unnecessary repeating of pattern matching is definitely less readable

Issue #402, opened 2023-04-13
[Tech]: Conditionally-public items for testing


bragov4ik:
### Problem statement

Make private items public only for test builds/runs. Only those items, that are actually needed in tests.

### Description

We are in process of changing our mocks to chainspec (#300). However, it introduced few problems (the change is tradeoff to some extent). For example, balances of accounts cannot be setup in [test externalities](https://github.com/paritytech/polkadot/blob/ae96e09e6ad1810b8d04d4530c07173e604a763d/xcm/pallet-xcm/src/tests.rs#L54).

One of such problems is that pallet tests cannot use private items. It happens because the parent module of tests (the pallet itself) is seen different from the pallet used in runtime (see [this issue](https://github.com/rust-lang/rust/issues/79381)). Therefore, we need to use pallet version from the runtime, for example `use framenode_runtime::order_book::Pallet;`. It is seen by Rust as separate module, disallowing to use private items.

There are 2 obvious ways to solve this:
1. Do not use private items in tests.
2. Make the items public and expose implementation details.

1 is often inconvenient, therefore we need to settle with 2, which may lead to unintentional usage in some other parts of the programs.

It makes sense to make them public only for test builds (`#[cfg(test)]`), so that the production code does not use the private items unintentionally.

### Definition of Done

1. Make a macro that allows such behaviour / use macro from external library
2. Use the macro with `#[cfg_attr(test, make_pub)]` where applicable

bragov4ik:
Had to use feature `"test"` because `#[cfg(test, visibility::make(pub))]` doesn't work. It happens because the crate with such attribute is treated as external in tests (basically it's a consequence of https://github.com/rust-lang/rust/issues/79381) and external crates are not built with `test` configuration.

For example, if `#[cfg(test, visibility::make(pub))]` is written in crate A and we have the same dependency case as in the issue, `test` will be propagated as follows:
```
A
|
B    A (test)
 \  /
  A (test)
```
and instance of A that is accessed through B is not built with `test` directive, which doesn't enable the attribute.

bragov4ik:
see #446

PR #446, opened 2023-05-15
402/test public

bragov4ik:


pallets/order-book/src/lib.rs
```diff
@@ -516,21 +518,21 @@ impl<T: Config> Pallet<T> {
         )
     }
 
-    // todo: make pub(tests) (k.ivanov)
     /// Validity of asset ids (for example, to have the same base asset
     /// for dex and pair) should be done beforehand
-    pub fn register_tech_account(
+    #[cfg_attr(feature = "test", visibility::make(pub))]
```
wer1st, line 523: Could we use something like this?
`#[cfg(test, visibility::make(pub))]`
In this case it is not necessary to use `test` feature falg

bragov4ik: Sorry, forgot to explain why I did this. Here's the explanation https://github.com/sora-xor/sora2-network/issues/402#issuecomment-1549344136

This approach is a bit ugly, I agree. If it's too bad I think we can leave it as-is and make everything public (features become a bit messy with `test`)

bragov4ik:
Decided to leave tested functions as public, because making a mandatory feature flag for `cargo test` could be unpleasant and frustrating; it should be able to run without flags.

PR #457, opened 2023-05-23
Exception for creation of pools involving XSTUSD

pakrentos:
# Changes
- If the XSTUSD is the base asset in the DEX, then the pool could be initialized

## Notes
LiquidityProxy with its current functionality allows swapping XSTUSD to some other assets directly, since it could be considered as base asset, so there was no need to modify it.

### DIFF
--- a/pallets/pool-xyk/src/lib.rs
+++ b/pallets/pool-xyk/src/lib.rs
@@ -808,7 +808,11 @@ pub mod pallet {
                     Error::<T>::UnableToCreatePoolWithIndivisibleAssets
                 );
 
-                let synthetic_assets = T::XSTMarketInfo::enabled_target_assets();
+                let mut synthetic_assets = T::XSTMarketInfo::enabled_target_assets();
+                // in case if XSTUSD is the base asset, we might want to create a pool anyways
+                // TODO: #392 use DexInfoProvider instead of dex-manager pallet
+                synthetic_assets
+                    .remove(&dex_manager::Pallet::<T>::get_dex_info(&dex_id)?.base_asset_id);
                 ensure!(
                     !synthetic_assets.contains(&asset_a) && !synthetic_assets.contains(&asset_b),
                     Error::<T>::UnableToCreatePoolWithSyntheticAssets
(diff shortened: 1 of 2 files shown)


vovac12 (approved):

pallets/pool-xyk/src/lib.rs
```diff
@@ -808,7 +808,11 @@ pub mod pallet {
                     Error::<T>::UnableToCreatePoolWithIndivisibleAssets
                 );
 
-                let synthetic_assets = T::XSTMarketInfo::enabled_target_assets();
+                let mut synthetic_assets = T::XSTMarketInfo::enabled_target_assets();
+                // in case if XSTUSD is the base asset, we might want to create a pool anyways
+                // TODO: #392 use DexInfoProvider instead of dex-manager pallet
+                synthetic_assets
+                    .remove(&dex_manager::Pallet::<T>::get_dex_info(&dex_id)?.base_asset_id);
```
vovac12, line 815: Seems like it will not deny creating XSTUSD-XOR and XSTUSD-XST pairs

vovac12 (changes requested):

runtime/src/lib.rs
```diff
@@ -952,6 +954,23 @@ impl technical::Config for Runtime {
...
+        (
+            0u32,
+            Box::new(
+                || <xst::Pallet::<Runtime> as GetMarketInfo<AssetId>>::enabled_target_assets()
+            ) as Box<dyn Fn() -> BTreeSet<AssetId>>
+        ),
+        (
+            1u32,
+            Box::new(|| {
+                let mut synthetics = <xst::Pallet::<Runtime> as GetMarketInfo<AssetId>>::enabled_target_assets();
```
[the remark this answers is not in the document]
pakrentos: This is the map only for target assets. In the place, where it is used, there is duplicate assets check in  `tech_account_from_dex_and_asset_pair`, so no need to remove XSTUSD from the map.

pallets/pool-xyk/src/lib.rs
```diff
@@ -681,6 +683,9 @@ pub mod pallet {
         type OnPoolReservesChanged: OnPoolReservesChanged<Self::AssetId>;
         /// Weight information for extrinsics in this pallet.
         type WeightInfo: WeightInfo;
+        type GetRestrictedTargetAssets: Get<
+            BTreeMap<Self::DEXId, Box<dyn Fn() -> BTreeSet<AssetIdOf<Self>>>>,
```
vovac12, line 687: You can replace it by `GetByKey` trait from orml traits

wer1st (approved):

vovac12 (approved):

PR #458, opened 2023-05-23
Order expiration

bragov4ik:


pallets/order-book/src/lib.rs
```diff
@@ -488,6 +554,8 @@ pub mod pallet {
             let dex_id = order_book.dex_id;
 
             order_book.cancel_limit_order::<Self>(order, &mut data)?;
+            Self::unschedule(expires_at, order_book_id, order_id)
+                .map_err(|e| <UnscheduleError as Into<Error<T>>>::into(e))?;
```
wer1st, line 558: Unscheduling should be move in `OrderBook` because it should be unscheduled when order is canceled or executed

wer1st: Btw, I would move `schedule()` in `OrderBook` as well to be consistent

pallets/order-book/src/limit_order.rs
```diff
@@ -72,7 +74,9 @@ impl<T: crate::Config + Sized> LimitOrder<T> {
         amount: OrderVolume,
         time: MomentOf<T>,
         lifespan: MomentOf<T>,
+        now: BlockNumberFor<T>,
```
wer1st, line 77:
```suggestion
        current_block: BlockNumberFor<T>,
```

pallets/order-book/src/lib.rs
```diff
@@ -592,6 +660,198 @@ impl<T: Config> Pallet<T> {
...
+    ExpirationNotFound,
+}
+
+impl<T: Config> Into<Error<T>> for UnscheduleError {
+    fn into(self) -> Error<T> {
+        match self {
+            UnscheduleError::ExpirationNotFound => Error::<T>::ExpirationNotFound,
+        }
+    }
+}
```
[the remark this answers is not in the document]
bragov4ik: (discussed in dms, replying for the record)
I thought that it would be nice to know that scheduling produces only a single error variant (especially it seemed useful in `on_initialize` part). But it turned out to be unused, so removing the types.

pallets/order-book/src/scheduler.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /// Expire orders that are scheduled to expire at `block`.
+    /// `weight` is used to track weight spent on the expirations, so that
+    /// it doesn't accidentally spend weight of the entire block (or even more).
+    ///
+    /// Returns `true` if all expirations were processed and `false` if some expirations
+    /// need to be retried when more weight is available.
+    pub fn service_block<TDataLayer: DataLayer<T>>(
+        data_layer: &mut TDataLayer,
```
wer1st, line 109:
```suggestion
    pub fn service_block(
        data_layer: &mut impl DataLayer<T>,
```

pallets/order-book/src/traits.rs
```diff
@@ -44,7 +44,7 @@ where
     fn get_limit_order(
         &mut self,
         order_book_id: &OrderBookId<AssetIdOf<T>>,
-        order_id: T::OrderId,
+        order_id: &T::OrderId,
```
wer1st, line 47: Not necessary to use ref

pallets/order-book/src/tests/order_book.rs
```diff
@@ -532,25 +573,37 @@ fn should_not_place_limit_order_that_doesnt_meet_restrictions_for_orders_in_pric
 
             buy_order.id += 1;
             buy_order.owner = account.clone();
+            // should ideally be set through `LimitOrder::new`
+            // but let's do it in a hacky way for simplicity
+            buy_order.expires_at += 1;
             sell_order.id += 1;
             sell_order.owner = account;
+            // should ideally be set through `LimitOrder::new`
+            // but let's do it in a hacky way for simplicity
```
wer1st, line 582: Comments are unnecessary

pallets/order-book/src/traits.rs
```diff
@@ -152,12 +151,33 @@ pub trait CurrencyUnlocker<AccountId, AssetId, DEXId> {
...
+        when: BlockNumber,
+        order_book_id: OrderBookId,
+        order_id: OrderId,
+    ) -> Result<(), Error>;
+    /// Remove the order from expiration schedule for block `when`.
+    fn unschedule(
+        when: BlockNumber,
+        order_book_id: OrderBookId,
+        order_id: OrderId,
+    ) -> Result<(), Error>;
```
wer1st, line 182:
```suggestion
    /// Execute scheduled expirations considering this block to be `current_block`
    /// and weight limit to be set by `weight`.
    ///
    /// If the weight limit is reached, it should continue where it's left at the
    /// next block.
    fn service(current_block: BlockNumber, weight: &mut WeightMeter);
    
    /// Schedule the order for expiration at block `when`.
    fn schedule(
        when: BlockNumber,
        order_book_id: OrderBookId,
        order_id: OrderId,
    ) -> Result<(), Error>;
    
    /// Remove the order from expiration schedule for block `when`.
    fn unschedule(
        when: BlockNumber,
        order_book_id: OrderBookId,
        order_id: OrderId,
    ) -> Result<(), Error>;
```

wer1st: Minor

pallets/order-book/src/tests/order_book.rs
bragov4ik: I intended to place them so that people that see this test don't use the same approach where it's not appropriate

vovac12 (approved):

PR #462, opened 2023-05-23
OrderBook: hidden market orders

wer1st:


pallets/order-book/src/lib.rs
```diff
@@ -314,6 +314,8 @@ pub mod pallet {
         Unauthorized,
         /// Invalid asset
         InvalidAsset,
+        /// Indicated limits for slippage has not been met during transaction execution.
```
bragov4ik, line 317:
```suggestion
        /// Indicated limit for slippage has not been met during transaction execution.
```

pallets/order-book/src/order_book.rs
```diff
@@ -261,12 +261,17 @@ impl<T: crate::Config + Sized> OrderBook<T> {
             *market_change.market_input.value(),
         )?;
 
+        let receiver = if let Some(to) = order.to {
+            to
+        } else {
+            order.owner
+        };
```
bragov4ik, line 268:
```suggestion
        let receiver = order.to.unwrap_or(order.owner);
```

pallets/order-book/src/lib.rs
```diff
@@ -644,37 +646,100 @@ impl<T: Config> LiquiditySource<T::DEXId, T::AccountId, T::AssetId, Balance, Dis
...
             )),
             QuoteAmount::WithDesiredOutput { .. } => Ok((
-                SwapOutcome::new(deal_info.input_amount, fee),
+                SwapOutcome::new(*deal_info.input_amount.value(), fee),
                 Self::quote_weight(),
             )),
         }
     }
 
     fn exchange(
```
bragov4ik, line 665: Just to clarify: the code that will call this function will be added later? Something like calling from liquidity proxy, right?

pallets/order-book/src/benchmarking.rs
```diff
@@ -486,10 +486,44 @@ benchmarks! {
...
+        .unwrap();
     }
     verify {
+        assert_eq!(
+            <T as Config>::AssetInfoProvider::free_balance(&order_book_id.base, &caller).unwrap(),
+            caller_base_balance - balance!(355.13473)
+        );
+        assert_eq!(
+            <T as Config>::AssetInfoProvider::free_balance(&order_book_id.quote, &caller).unwrap(),
+            caller_quote_balance + balance!(3499.999935)
```
bragov4ik, line 525: Maybe use `assert_approx_eq` if it's due to FP math

bragov4ik (approved):
LGTM

bragov4ik (approved):
LGTM

pallets/order-book/src/benchmarking.rs
wer1st: I prefer to work with exact numbers

pallets/order-book/src/lib.rs
wer1st: It is already added in #444 
It is called from `liquidity-proxy`

pallets/order-book/src/benchmarking.rs
bragov4ik: It just seems like some magic number

pallets/order-book/src/lib.rs
bragov4ik: Oh, ok. Couldn't find it at first because it uses macro `exchange!` instead of usual function call.
Thanks, I see it now

vovac12 (approved):

Issue #400, opened 2023-04-13
OrderBook. CRUD operations permissions


bgodlin:
<table class="c20">
      <tr class="c13">
        <td class="c12" colspan="2" rowspan="2">
          <h1 class="c21" id="h.m9vhjvp9v18w">
            <span class="c0">&#9196; Asset type</span>
          </h1>
        </td>
        <td class="c24" colspan="3" rowspan="1">
          <h1 class="c21" id="h.swtin4vpkm0k"><span class="c16">CRUD</span></h1>
        </td>
      </tr>
      <tr class="c13">
        <td class="c3" colspan="1" rowspan="1">
          <h1 class="c21" id="h.swtin4vpkm0k-6">
            <span class="c0">Creation</span>
          </h1>
        </td>
        <td class="c6" colspan="1" rowspan="1">
          <h1 class="c21" id="h.swtin4vpkm0k-7">
            <span class="c0">Parameter modification</span>
          </h1>
        </td>
        <td class="c6" colspan="1" rowspan="1">
          <h1 class="c21" id="h.h3rggpr2qcmr">
            <span class="c0">Status modification and removal</span>
          </h1>
        </td>
      </tr>
      <tr class="c15">
        <td class="c11" colspan="1" rowspan="2">
          <p class="c10"><span class="c9">Regular asset</span></p>
        </td>
        <td class="c4" colspan="1" rowspan="1">
          <p class="c10"><span class="c9"><a class="c25"
                href="https://github.com/sora-xor/polkaswap-token-whitelist-config/blob/master/whitelist.json"
                >Default</a></span></p>
        </td>
        <td class="c3" colspan="1" rowspan="1">
          <p class="c10">
            <span class="c8"
              >Automatically on runtime upgrade from the whitelisted technical
              account</span
            >
          </p>
          <p class="c10"> Default parameters:
            <li class="c5 li-bullet-0"><span class="c8">Price step: 0.00001 XOR</span></li>
            <li class="c5 li-bullet-0"><span class="c8">Volume step: 0.00001</span></li>
          </p>
        </td>
        <td class="c6" colspan="1" rowspan="1">
          <ul class="c18 lst-kix_gnd1hip7nte-0 start">
            <li class="c5 li-bullet-0"><span class="c8">Asset owner</span></li>
            <li class="c5 li-bullet-0"><span class="c8"><a
                href="https://github.com/sora-xor/sora2-network/issues/413"
[...]

PR #468, opened 2023-05-26
Change order book status

wer1st:


vovac12 (changes requested):
Currently status can be changed only by root, but tech committee should also be able to do it according to the task
You should add `UpdateOrigin: EnsureOrigin<Self::RuntimeOrigin>` to `Config` trait and specify it like this:
```rust
type UpdateOrigin = EitherOfDiverse<
        pallet_collective::EnsureProportionMoreThan<AccountId, TechnicalCollective, 1, 2>,
        EnsureRoot<AccountId>,
    >;
```

pallets/order-book/src/lib.rs
```diff
@@ -432,12 +439,21 @@ pub mod pallet {
...
+            status: OrderBookStatus,
         ) -> DispatchResult {
             ensure_root(origin)?;
-            // todo (m.tagirov)
-            todo!()
+            let mut order_book =
+                <OrderBooks<T>>::get(order_book_id).ok_or(Error::<T>::UnknownOrderBook)?;
+            let dex_id = order_book.dex_id;
+            order_book.status = status;
+            <OrderBooks<T>>::set(order_book_id, Some(order_book));
```
bragov4ik, line 450:
```suggestion
            let dex_id = <OrderBooks<T>>::mutate(order_book_id, |order_book| {
                let order_book = order_book.as_mut().ok_or(Error::<T>::UnknownOrderBook)?;
                order_book.status = status;
                Ok::<_, Error<T>>(order_book.dex_id)
            })?;
```

pallets/order-book/src/benchmarking.rs
```diff
@@ -351,10 +355,30 @@ benchmarks! {
     }
 
     change_orderbook_status {
+        let order_book_id = OrderBookId::<AssetIdOf<T>> {
+            base: VAL.into(),
+            quote: XOR.into(),
+        };
+
+        create_and_fill_order_book::<T>(order_book_id);
```
bragov4ik, line 363: Maybe use `create_empty_order_book` instead (this benchmark doesn't seem to be affected by placed orders)?

Also I noticed that there are two `create_and_fill_order_book` functions that don't seem to be different (correct me if I'm wrong); maybe we should remove one of them?

bragov4ik (approved):
LGTM (apart from permissions which were mentioned before)

wer1st:
> Currently status can be changed only by root, but tech committee should also be able to do it according to the task You should add `UpdateOrigin: EnsureOrigin<Self::RuntimeOrigin>` to `Config` trait and specify it like this:
> 
> ```rust
> type UpdateOrigin = EitherOfDiverse<
>         pallet_collective::EnsureProportionMoreThan<AccountId, TechnicalCollective, 1, 2>,
>         EnsureRoot<AccountId>,
>     >;
> ```

We will do that in the separate PR for task #400

pallets/order-book/src/benchmarking.rs
wer1st: 1. It looks better when we benchmark it with filled order book.
2. We have it in `benchmarking.rs` & `test_utils.rs`. Benchmarking versrion depends on T, the test version depends directly on Runtime. The idea is that benchmarks could be run as benchmarks and as tests. In order book we could use it only as benchmarks and cannot as tests for a while. If we want to use only 1 function we should move it out o the test scope. It wil look like test util outside test module. I think we could decide what to do with this function later when we have a new solution for benchmarks/tests/mocks or write some playground pallet for QA. After that it could be more clear how we can rework it.

vovac12 (approved):

PR #512, opened 2023-06-01
Recovery pallet added and configured

bgodlin:


vovac12 (approved):

runtime/src/lib.rs
```diff
@@ -2012,6 +2012,25 @@ impl order_book::Config for Runtime {
     type WeightInfo = order_book::weights::SubstrateWeight<Runtime>;
 }
 
+parameter_types! {
+
```
wer1st, line 2016: redundant empty line

runtime/Cargo.toml
```diff
@@ -249,6 +251,7 @@ std = [
     "pallet-offences/std",
     "pallet-preimage/std",
     "pallet-randomness-collective-flip/std",
+    "pallet-recovery/std",
```
wer1st, line 254: Probably we need to add `"pallet-recovery/try-runtime"` below if it has this feature

runtime/src/lib.rs
bgodlin: Resolved in https://github.com/sora-xor/sora2-network/pull/512/commits/50e06c86051720b8bde516dbbb2abf2a474926eb

runtime/Cargo.toml
bgodlin: I think it works without it

wer1st: Of course it works, but it is necessary for cases when we use `try-runtime`

Alexey-N-Chernyshov (commented):
Is it relevant?

bgodlin:
> Is it relevant?

It is

WRRicht3r (approved):
This would be an interesting feature to have for n00b users to have less fear about self-custodial. Let's consider implementing as an "optional extra."

runtime/src/lib.rs
```diff
@@ -1967,6 +1967,24 @@ impl order_book::Config for Runtime {
...
+parameter_types! {
+    pub const ConfigDepositBase: Balance = balance!(5);
+    pub const FriendDepositFactor: Balance = balance!(3);
+    pub const MaxFriends: u16 = 10;
+    pub const RecoveryDeposit: Balance = balance!(5);
+}
+
+impl pallet_recovery::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type WeightInfo = pallet_recovery::weights::SubstrateWeight<Runtime>;
```
bragov4ik, line 1979: I think we're supposed to benchmark it ourselves (with our reference hardware)

PR #529, opened 2023-06-07
Transformation of Limit order into Market order when crossing the spread

wer1st:


pallets/order-book/src/lib.rs
```diff
@@ -277,23 +277,41 @@ pub mod pallet {
...
+        /// User tried to place the limit order out of the spread. The limit order is converted into a market order.
+        LimitOrderConvertedToMarketOrder {
+            order_book_id: OrderBookId<AssetIdOf<T>>,
+            dex_id: T::DEXId,
+            owner_id: T::AccountId,
+        },
+
+        /// User tried to place the limit order out of the spread.
+        /// One part of the liquidity of the limit order is converted into a market order, and the other part is placed as a limit order.
+        LimitOrderSplittedIntoMarketOrderAndLimitOrder {
```
bragov4ik, line 296:
```suggestion
        LimitOrderSplitIntoMarketOrderAndLimitOrder {
```
just caught my eye :)

pallets/order-book/src/types.rs
```diff
@@ -186,11 +224,169 @@ impl<AssetId: PartialEq> DealInfo<AssetId> {
...
+        }
+    }
+
+    pub fn merge(&mut self, other: &Self) -> Result<(), ()> {
+        ensure!(self.dex_id == other.dex_id, ());
+        ensure!(self.order_book_id == other.order_book_id, ());
+
+        let merge_asset_map =
+            |map: &mut BTreeMap<AssetId, BTreeMap<AccountId, OrderVolume>>,
+             to_merge: &BTreeMap<AssetId, BTreeMap<AccountId, OrderVolume>>| {
```
bragov4ik, line 255: I got some typescript flashbacks :) Usual `fn` should look less complicated + will have one less nesting level
What do you think?

pallets/order-book/src/types.rs
```diff
@@ -186,11 +224,169 @@ impl<AssetId: PartialEq> DealInfo<AssetId> {
...
+                to_merge.iter().for_each(|(asset, whom)| {
```
bragov4ik, line 256: Maybe use simple `for`? Should look cleaner, what do you think?

pallets/order-book/src/types.rs
```diff
@@ -186,11 +224,169 @@ impl<AssetId: PartialEq> DealInfo<AssetId> {
...
+        merge_asset_map(&mut self.to_unlock, &other.to_unlock);
+
+        Ok(())
+    }
+
+    pub fn lock<Locker>(&self) -> Result<(), DispatchError>
+    where
+        Locker: CurrencyLocker<AccountId, AssetId, DEXId, DispatchError>,
+    {
+        for (asset_id, from_whom) in self.to_lock.iter() {
```
bragov4ik, line 280:
```suggestion
        for (asset_id, from_whom) in self.to_lock {
```
afair `.iter()` is automatically called in for loops

pallets/order-book/src/types.rs
```diff
@@ -186,11 +224,169 @@ impl<AssetId: PartialEq> DealInfo<AssetId> {
...
+            to_add: BTreeMap::new(),
+            to_update: BTreeMap::new(),
+            to_delete: BTreeMap::new(),
+            payment: Payment::new(dex_id, order_book_id),
+            ignore_unschedule_error: false,
+        }
+    }
+
+    pub fn merge(&mut self, mut other: Self) -> Result<(), ()> {
+        let join = |lhs: Option<OrderAmount>,
```
bragov4ik, line 364: also maybe just `fn`

pallets/order-book/src/order_book.rs
```diff
@@ -361,23 +497,43 @@ impl<T: crate::Config + Sized> OrderBook<T> {
...
         ensure!(remaining_amount.is_zero(), Error::<T>::NotEnoughLiquidity);
 
-        let (market_input, market_output) = match side {
+        payment
+            .to_lock
+            .entry(maker_out_asset)
+            .or_default()
+            .entry(taker)
+            .and_modify(|lock_amount| *lock_amount += maker_amount)
+            .or_insert(maker_amount);
```
bragov4ik, line 506: Might be handy to make functions like `lock` in `Payment` (and rename existing `lock` to `execute_lock`) and just pass three parameters instead of writing this long expression
not critical ofc

pallets/order-book/src/order_book.rs
```diff
@@ -688,8 +874,93 @@ impl<T: crate::Config + Sized> OrderBook<T> {
...
+            T::DEXId,
+            T::OrderId,
+            LimitOrder<T>,
+            BlockNumberFor<T>,
+        >,
+        DispatchError,
+    > {
+        let (market_amount, limit_amout) = match limit_order.side {
+            PriceVariant::Buy => Self::calculate_market_depth_to_price(
+                limit_order.side.switch(),
```
bragov4ik, line 894: Not an addition of this PR but name `switched()` should be more accurate

pallets/order-book/src/order_book.rs
```diff
@@ -688,8 +874,93 @@ impl<T: crate::Config + Sized> OrderBook<T> {
...
+            new_limit_order.amount = limit_amout;
+            market_change
+                .merge(self.calculate_limit_order_impact(new_limit_order)?)
+                .map_err(|_| Error::<T>::AmountCalculationFailed)?;
+        }
+
+        Ok(market_change)
+    }
+
+    /// Calculates and returnes the sum of limit orders up to the `price`or until the `amount` is reached
```
bragov4ik, line 931:
```suggestion
    /// Calculates and returns the sum of limit orders up to the `price` or until the `amount` is reached
```

pallets/order-book/src/order_book.rs
```diff
@@ -688,8 +874,93 @@ impl<T: crate::Config + Sized> OrderBook<T> {
...
+    /// and remainig `amount` if it is greater then the market volume.
```
bragov4ik, line 932:
```suggestion
    /// and remaining `amount` if it is greater than the market volume.
```

bragov4ik (approved):
LGTM, only minor fixes

pallets/order-book/src/types.rs
```diff
@@ -186,11 +224,169 @@ impl<AssetId: PartialEq> DealInfo<AssetId> {
...
+#[derive(Eq, PartialEq, Clone, RuntimeDebug)]
+pub struct MarketChange<AccountId, AssetId, DEXId, OrderId, LimitOrder, BlockNumber> {
+    // Info fields
+    /// The amount of the input asset for the exchange deal
+    pub deal_input: Option<OrderAmount>,
+
+    /// The amount of the output asset for the exchange deal
+    pub deal_output: Option<OrderAmount>,
+
+    /// The amount of the input asset that is placed into the market
```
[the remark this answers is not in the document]
wer1st: Yes, as limit orders.
Order book with limit orders is a market.
Market order is a deal to exchange assets in this market

pallets/order-book/src/lib.rs
```diff
@@ -277,23 +277,41 @@ pub mod pallet {
...
             dex_id: T::DEXId,
             order_id: T::OrderId,
             owner_id: T::AccountId,
         },
 
+        /// User tried to place the limit order out of the spread. The limit order is converted into a market order.
+        LimitOrderConvertedToMarketOrder {
+            order_book_id: OrderBookId<AssetIdOf<T>>,
+            dex_id: T::DEXId,
+            owner_id: T::AccountId,
```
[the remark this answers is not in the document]
wer1st: Limit order could be placed only be extrinsic. All extrinsics are executed one by one and associated event is emited in the scope of extrinsic executing.

pallets/order-book/src/order_book.rs
wer1st: Disagree. `lock` is a verb and it is logical that assets are locked immediately when you call `lock` method

bragov4ik: Makes sense, then something like `schedule_lock` should fit

pallets/order-book/src/order_book.rs
wer1st: `switch` as a verb in the imperative mood is more accurate

pallets/order-book/src/types.rs
```diff
@@ -186,11 +224,169 @@ impl<AssetId: PartialEq> DealInfo<AssetId> {
...
+    pub fn new(dex_id: DEXId, order_book_id: OrderBookId<AssetId>) -> Self {
+        Self {
+            dex_id,
+            order_book_id,
+            to_lock: BTreeMap::new(),
+            to_unlock: BTreeMap::new(),
+        }
+    }
+
+    pub fn merge(&mut self, other: &Self) -> Result<(), ()> {
```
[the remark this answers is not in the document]
wer1st: Ideally, it should never happen. However there is a more descriptive erron on the level above where this method is called.

pallets/order-book/src/types.rs
wer1st: No difference

pallets/order-book/src/types.rs
wer1st: I use here Entry API & HOF approach. It is ok

pallets/order-book/src/types.rs
wer1st: The logic is not complicated, the closure is suitable. Also if we move it to the separate function, it becomes possible to reuse it, but I want to avoid it because it is merge-specific behavior

pallets/order-book/src/order_book.rs
bragov4ik: however it does not change the state of `self` but rather returns a new value (that is actually switched)

pallets/order-book/src/types.rs
wer1st: No, `.into_iter()` is automatically called in for loops. And it consumes the collection

pallets/order-book/src/order_book.rs
wer1st: Ok, but it is not my code =)

pallets/order-book/src/types.rs
```diff
@@ -186,11 +224,169 @@ impl<AssetId: PartialEq> DealInfo<AssetId> {
...
+                                    .entry(account.clone())
+                                    .and_modify(|current_volune| *current_volune += volume)
+                                    .or_insert(*volume);
+                            })
+                        })
+                        .or_insert(whom.clone());
+                });
+            };
+
+        merge_asset_map(&mut self.to_lock, &other.to_lock);
```
vovac12, line 270: Maybe use something like this here?
```rust
for (map, to_merge) in &[
  (&mut self.to_lock, &other.to_lock), 
  (&mut self.to_unlock, &other.to_unlock)
] {
  ...
}

vovac12 (approved):
LGTM

pallets/order-book/src/types.rs
bragov4ik: There are no higher-order functions here and `fn` still allows using `entry `

pallets/order-book/src/types.rs
bragov4ik: Then `for ... in &self.lock` should work
But ok, here it's rather a matter of preference

pallets/order-book/src/types.rs
bragov4ik: Closure adds more unnecessary syntax here and I meant to define `fn` in the scope of merge [like so](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=db46f931086de704a2e89e40383744f4)

pallets/order-book/src/types.rs
bragov4ik: Also splitting it into two functions can make reading it easier:
```rust
    fn merge_account_map(
        account_map: &mut BTreeMap<AccountId, OrderVolume>,
        to_merge: &BTreeMap<AccountId, OrderVolume>
    ) {
        for (account, volume) in to_merge {
            account_map
                .entry(account.clone())
                .and_modify(|current_volune| *current_volune += volume)
                .or_insert(*volume);
        }
    }

    fn merge_asset_map(
        map: &mut BTreeMap<AssetId, BTreeMap<AccountId, OrderVolume>>,
        to_merge: &BTreeMap<AssetId, BTreeMap<AccountId, OrderVolume>>
    ) {
        for (asset, whom) in to_merge {
            map.entry(*asset)
                .and_modify(|account_map| {
                    merge_account_map(account_map, whom);
                })
                .or_insert(whom.clone());
        }
    }
 ```

Alternatively, some comment explaining what it does should help

PR #532, opened 2023-06-07
Hermes Governance platform upgrade

DanijelCBS:
Changed poll options logic and minimum voting period


pallets/hermes-governance-platform/src/lib.rs
```diff
@@ -46,21 +42,32 @@ pub struct HermesPollInfo<AccountId, Moment> {
...
     /// Poll title
-    pub title: String,
+    pub title: BoundedString<StringLimit>,
     /// Description
-    pub description: String,
+    pub description: BoundedString<StringLimit>,
     /// Creator Hermes withdrawn
     pub creator_hermes_withdrawn: bool,
+    /// Options
+    pub options: Vec<BoundedString<StringLimit>>,
```
vovac12, line 51: Maybe also use BoundedVec instead of Vec?

vovac12 (commented):
LGTM, but it would be nice to replace `Vec` with `BoundedVec`

pallets/hermes-governance-platform/src/lib.rs
JovanCBS: Done.

vovac12 (approved):

PR #552, opened 2023-06-15
Allow only Polkaswap DEX for OrderBook creation

bragov4ik:
Check the dex_id + minor fixes:
- rename `switch` to `switched` to more accurately represent meaning of the function
- fix heavy tests to surpass scheduler limits

### DIFF
--- a/pallets/order-book/src/lib.rs
+++ b/pallets/order-book/src/lib.rs
@@ -130,6 +130,7 @@ pub mod pallet {
         type MaxSidePriceCount: Get<u32>;
         type MaxExpiringOrdersPerBlock: Get<u32>;
         type MaxExpirationWeightPerBlock: Get<Weight>;
+        type AllowedDEXIds: Get<&'static [Self::DEXId]>;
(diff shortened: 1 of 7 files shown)


pallets/order-book/src/lib.rs
```diff
@@ -414,6 +416,11 @@ pub mod pallet {
                 order_book_id.base != order_book_id.quote,
                 Error::<T>::ForbiddenToCreateOrderBookWithSameAssets
             );
+            // https://github.com/sora-xor/sora2-network/issues/536
```
wer1st, line 419: comment is redundamt

wer1st (approved):

pallets/order-book/src/lib.rs
bragov4ik: I thought it could explain why this check is needed (it's not obvious for a person without a context of this problem)
Now that I think about it, `git blame` should show this PR, so it should be fine without the comment, thanks :)

vovac12 (approved):

PR #581, opened 2023-06-29
Testing tools (qa-tools pallet)

bragov4ik:
Private-net-only pallet with two extrinsics. Aimed to help testing process on a running blockchain.


pallets/qa-tools/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+                    now,
+                )?;
+            }
+            data.commit();
+            Ok(())
+        }
+
+        /// Fill a single order book.
+        fn fill_order_book(
+            data: &mut CacheDataLayer<T>,
```
wer1st, line 290:
```suggestion
            data: &mut impl DataLayer<T>,
```

wer1st: We could use the trait to make it more "universal". The same below

pallets/qa-tools/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+                .iter()
+                .map(|(quote, base)| {
+                    let quote_amount_fixed = FixedWrapper::from(*quote) * FixedWrapper::from(*base);
+                    quote_amount_fixed.into_balance()
+                })
+                .sum();
+            let sell_base_locked: Balance = sell_orders.iter().map(|(_, base)| base).sum();
+
+            // mint required amount to make this extrinsic self-sufficient
+            assets::Pallet::<T>::mint_unchecked(&book_id.quote, &bids_owner, buy_quote_locked)?;
```
wer1st, line 331: Usually we use update_balance

bragov4ik: `update_own_balance` requires mint permissions for the account, the extrinsic (`update_balance`) is undesirable to call, and I recall we wanted to use `assets` pallet as a single interface to operate with the balances

I'm not sure how to do it otherwise

wer1st: ok

Cargo.lock
```diff
@@ -27,7 +27,7 @@ version = "0.19.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a76fd60b23679b7d19bd066031410fb7e458ccc5e958eb5c325888ce4baedc97"
 dependencies = [
- "gimli 0.27.3",
+ "gimli 0.27.2",
```
wer1st, line 30: Why older versions? Here and below

pallets/order-book/src/lib.rs
```diff
@@ -457,31 +457,11 @@ pub mod pallet {
...
         ) -> DispatchResult {
             let who = ensure_signed(origin)?;
-            ensure!(
-                order_book_id.base != order_book_id.quote,
-                Error::<T>::ForbiddenToCreateOrderBookWithSameAssets
-            );
             ensure!(
                 order_book_id.dex_id == common::DEXId::Polkaswap.into(),
                 Error::<T>::NotAllowedDEXId
             );
```
wer1st, line 463: why didn't you move it into `verify_create_orderbook_params`?

bragov4ik: I'm not certain that it shouldn't be in `verify_create_orderbook_params`, so I can move it here if necessary

Cargo.lock
bragov4ik: Good question, let me check

pallets/order-book/src/lib.rs
wer1st: It is a part of this check

pallets/qa-tools/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+    impl<T: Config> Pallet<T> {
+        /// Create multiple order books with default parameters (if do not exist yet).
+        ///
+        /// Parameters:
+        /// - `origin`: caller, should be account because error messages for unsigned txs are unclear,
+        /// - `order_book_ids`: ids of the created order books; trading pairs are created
+        /// if necessary,
+        #[pallet::call_index(0)]
+        #[pallet::weight(<T as Config>::WeightInfo::order_book_create_empty_many())]
+        pub fn order_book_create_empty_many(
```
wer1st, line 121:
```suggestion
        pub fn create_empty_order_books_batch(
```

pallets/qa-tools/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+            Ok(PostDispatchInfo {
+                actual_weight: None,
+                pays_fee: Pays::No,
+            })
+        }
+    }
+
+    impl<T: Config> Pallet<T> {
+        /// Does not create an order book if it already exists
+        fn create_multiple_empty_unchecked(
```
wer1st, line 202:
```suggestion
        fn create_multiple_empty_order_books_unchecked(
```

pallets/qa-tools/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+        /// - `origin`: caller, should be account because unsigned error messages are unclear,
+        /// - `dex_id`: DEXId for all created order books,
+        /// - `bids_owner`: Creator of the buy orders placed on the order books,
+        /// - `asks_owner`: Creator of the sell orders placed on the order books,
+        /// - `fill_settings`: Parameters for placing the orders in each order book.
+        /// `best_bid_price` should be at least 3 price steps from the lowest accepted price,
+        /// and `best_ask_price` - at least 3 steps below maximum price,
+        #[pallet::call_index(1)]
+        #[pallet::weight(<T as Config>::WeightInfo::order_book_create_and_fill_many())]
+        pub fn order_book_create_and_fill_many(
```
wer1st, line 163:
```suggestion
        pub fn create_and_fill_order_books_batch(
```

pallets/qa-tools/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+            }
+            Ok(())
+        }
+
+        /// Place orders into the orderbook.
+        ///
+        /// In fill settings, `best_bid_price` should be at least 3 price steps from the
+        /// lowest accepted price, and `best_ask_price` - at least 3 steps below
+        /// maximum price.
+        fn fill_multiple_empty_unchecked(
```
wer1st, line 243:
```suggestion
        fn fill_multiple_empty_order_books_unchecked(
```

runtime/src/lib.rs
```diff
@@ -1586,6 +1588,17 @@ impl faucet::Config for Runtime {
     type WeightInfo = faucet::weights::SubstrateWeight<Runtime>;
 }
 
+parameter_types! {
+    pub QaToolsOrderLifespan: Moment = 60*10*1000; // 10 minutes
```
wer1st, line 1592: 10 mins is too small

pallets/qa-tools/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+                        order_book_id.quote.into(),
+                        order_book_id.base.into(),
+                    )?;
+                }
+                if <T as pallet::Config>::AssetInfoProvider::total_balance(
+                    &order_book_id.base,
+                    &who,
+                )? == Balance::zero()
+                {
+                    assets::Pallet::<T>::mint_unchecked(&order_book_id.base, &who, 1)?;
```
wer1st, line 227: For what?

wer1st: Btw, your test is broken by this.

bragov4ik: I forgot to add that it's only for nft(non-div assets)

pallets/qa-tools/src/lib.rs
bragov4ik: The idea is to separate all extrinsics for order books to distinguish them with addition of other callables into this pallet

pallets/qa-tools/src/lib.rs
bragov4ik: restructured the pallet

wer1st (approved):

pallets/qa-tools/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+                    pays_fee: Pays::No,
+                },
+                error: e,
+            })?;
+
+            // Extrinsic is only for testing, so we return all fees
+            // for simplicity.
+            Ok(PostDispatchInfo {
+                actual_weight: None,
+                pays_fee: Pays::No,
```
vovac12, line 131: I don't think it's good idea to have such extrinsics on stage or even test environment. We can mint as many tokens on test environments as we want for testing, but some bad users can try to break our testnets with this extrinsics. Another way to avoid such problem is to make this extrinsics callable only by root or trusted account

pallets/qa-tools/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+            BalancePrecision,
+            ContentSource,
+            Description,
+        >;
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
+        /// Order book related errors
+        OrderBook(pallets::order_book_tools::Error),
```
vovac12, line 76: Does it work fine with polkadot.js? And how it is represented in metadata?

pallets/qa-tools/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    use order_book::{MomentOf, OrderBookId};
+    use sp_std::prelude::*;
+
+    #[pallet::pallet]
+    pub struct Pallet<T>(_);
+
+    #[pallet::config]
+    pub trait Config: frame_system::Config + order_book::Config + trading_pair::Config {
+        type WeightInfo: WeightInfo;
+        type OrderBookOrderLifespan: Get<MomentOf<Self>>;
```
vovac12, line 61: `order_book::Config` trait already contains the same constant, why you need to duplicate it?

bragov4ik: Constants in `order_book` specify boundaries for the order lifespans, but this one dictates lifetimes of orders created with this pallet (value instead of the allowed range)

Now that I think about it, maybe we should specify the lifetime in extrinsic call instead.. @wer1st what do you think?

pallets/qa-tools/src/lib.rs
bragov4ik: I'll check it

pallets/qa-tools/src/lib.rs
wer1st: Yes, we could move it to extrinsics

pallets/qa-tools/src/lib.rs
bragov4ik: Since we try to "record" the development process with comments on issues (also why not PRs), @ra9mls, can you clarify please?)

pallets/qa-tools/src/lib.rs
bragov4ik: Ok, will do

pallets/qa-tools/src/lib.rs
bragov4ik: It works but doesn't show the underlying error :( 
image
image

bragov4ik: only `dispatchError.Module.error` field in polkadot.js distinguishes between underlying variants:
image
image

bragov4ik: So it seems like not the best solution in terms of polkadot.js user experience.

I wanted to somehow show that the error is related to order book pallet. The ways to do it that I see:

- Prefix error names with `OrderBook`, like `OrderBookUnknownOrderBook` _- seems kinda clumsy and inconvenient_
- Group them only in code - _not exposed externally (in particular to polkadot.js), so the users (testers in this case) should be expected to understand which pallet it refers to from context (might be ok considering that the error occurs right after the call)_
image
- Use existing error from `order_book` pallet (fits for this particular case, not sure about future versions of the pallet) - _should correctly display in polkadot.js, clearly separated in code, reuses existing errors (not sure if it's good or bad)_

bragov4ik: I lean towards prefixing in general, but reusing errors should be the best option in this particular case.

what do you think @vovac12 @wer1st ?

wer1st: @vovac12 can we have errors with fields? Such as `QaToolError(OrderBookError)` or `QaToolError(String)`?

vovac12: @bragov4ik you can use error names without prefix, group them in code and add detailed error description in doc comment

@wer1st no, dispatch error can only contain pallet index and error index

pallets/qa-tools/src/lib.rs
vovac12: I think we can add storage with whitelisted accounts to this pallet and add new accounts to whitelist via root exrinsic

vovac12: It's good to have weights for this extrinsics, but it's not so big problem if we will have restricted access to this pallet

bragov4ik: > I think we can add storage with whitelisted accounts to this pallet and add new accounts to whitelist via root exrinsic

Seems good! Just want to clarify, @ra9mls, is this approach convenient from QA perspective?

bragov4ik: Got the answer in DM, it's ok.

pallets/qa-tools/src/lib.rs
bragov4ik: Ok, seems like it's the best option in this case

vovac12 (approved):

wer1st (approved):

PR #587, opened 2023-07-03
Fix/xst limit incorporate fees

pakrentos:
# Changes
- Fees are now taken into account during XST limit check (Selling XST with desired output case)
- Added test for checking quotes with amounts near limits
- Added checks and tests on 0 input/output amount
- Fixed oracle dynamic fee calculation + rechecked maths + added tests
- Added check on resulting fee ratio in XST (should be in bounds [0;1))


pallets/xst/src/lib.rs
```diff
@@ -64,6 +64,7 @@ use frame_support::{ensure, fail, RuntimeDebug};
 use serde::{Deserialize, Serialize};
 use sp_runtime::DispatchError;
 use sp_std::collections::btree_set::BTreeSet;
+use sp_std::if_std;
```
wer1st, line 67: unused

wer1st (approved):

pallets/band/src/lib.rs
```diff
@@ -541,7 +541,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
         let decayed_fee = prev_fee * decay;
 
         let new_fee_part =
-            (prev_rate / new_rate - fixed_wrapper!(1)).abs() - doubled_deviation - min_fee;
+            (new_rate / prev_rate - fixed_wrapper!(1)).abs() - doubled_deviation - min_fee;
         let new_fee = if new_fee_part > fixed_wrapper!(0) {
             decayed_fee + new_fee_part
```
wer1st, line 546: You use `abs()` here. But what if `new_rate` is less than `prev_rate`? In this case we still increase `new_fee`. Is it correct?

pakrentos: Good question, it may be a good idea to make the fee only grow with pumps, since we can't short the synthetics in SORA (at least now). But I'm not sure if dumps do not make positive influence during long high volatility periods. Anyway, sounds good, will fix this

PR #614, opened 2023-07-13
Cancel limit orders batch

wer1st:


pallets/order-book/src/lib.rs
```diff
@@ -765,6 +765,32 @@ pub mod pallet {
             data.commit();
             Ok(().into())
         }
+
+        #[pallet::call_index(6)]
+        #[pallet::weight(<T as Config>::WeightInfo::cancel_limit_order().saturating_mul(limit_orders_to_cancel.iter().fold(0, |count, (_, order_ids)| count + order_ids.len() as u64)))]
```
bragov4ik, line 770:
```suggestion
        #[pallet::weight(<T as Config>::WeightInfo::cancel_limit_order().saturating_mul(limit_orders_to_cancel.iter().map(|(_, order_ids)| order_ids.len() as u64).sum()))]
```

pallets/order-book/src/traits.rs
```diff
@@ -123,12 +123,21 @@ where
...
-    /// Returns order ids of user
+    /// Returns order ids of user from the order book with `order_book_id`
     fn get_user_limit_orders(
         &mut self,
         account: &T::AccountId,
         order_book_id: &OrderBookId<AssetIdOf<T>, T::DEXId>,
     ) -> Option<UserOrders<T::OrderId, T::MaxOpenedLimitOrdersPerUser>>;
+
+    /// Returns order ids of user from all order books
+    fn get_all_user_limit_orders(
```
bragov4ik, line 134: is this method going to be used later? just want to clarify that it's needed

bragov4ik (approved):
lgtm

pallets/order-book/src/lib.rs
wer1st: I prefer fold

pallets/order-book/src/traits.rs
wer1st: It was used for previos solution until we decided to change the approach. It could be useful for some cases and I decided to not remove it

PR #649, opened 2023-08-01
Market orders for non-divisible assets

wer1st:


vovac12 (approved):

pallets/order-book/src/lib.rs
```diff
@@ -449,8 +449,8 @@ pub mod pallet {
         MaxLotSizeIsMoreThanTotalSupply,
         /// Indicated limit for slippage has not been met during transaction execution.
         SlippageLimitExceeded,
-        /// NFT order books are temporarily forbidden
-        NftOrderBooksAreTemporarilyForbidden,
+        /// Market orders are allowed only for indivisible assets
```
bragov4ik, line 452: Maybe write that instead it should be done through `liquidity_proxy` and namely `LiquiditySource::exchange` for divisible assets? Otherwise it seems quite confusing.
Something like
```suggestion
        /// Market orders are allowed only for indivisible assets.
        /// Divisible assets are traded through `swap` from `liquidity_proxy` pallet, which uses
        /// `LiquiditySource::exchange`.
```

bragov4ik (approved):

pallets/order-book/src/lib.rs
wer1st: User should't know about internal `LiquiditySource::exchange`.  And there is a documentation where it is already should be described.

bragov4ik: > And there is a documentation where it is already should be described.

Where is it? 🤔

PR #660, opened 2023-08-07
XST Fallback mechanism

pakrentos:
# Changes
## XST
- Introduced new extrinsic: XSTPool::remove_synthetic_asset. Called by root, used deleting synthetic asset entirely
- XSTPool::disable_synthetic_asset now only removes info from EnabledSynthetics, keeping the EnabledSymbols.
- XSTPool doesn't care about 5 min limit on symbol rates - now using only Oracle::quote_unchecked
- New storage map in Band: SymbolCheckBlock(block_number, symbol -> bool). Holds info about when the symbol should be disabled in case of no update (via calling OnSymbolDisabledHook::disable_symbol)


vovac12 (approved):

pallets/xst/src/lib.rs
```diff
@@ -1005,6 +1028,11 @@ impl<T: Config> Pallet<T> {
             )),
         }
     }
+
+    fn disable_synthetic_asset_unchecked(synthetic_asset: AssetIdOf<T>) {
+        EnabledSynthetics::<T>::remove(synthetic_asset);
```
wer1st, line 1033: Now you can also disable XSTPool source for the trading pair by `trading_pair::disable_source_for_trading_pair`

pallets/xst/src/mock.rs
```diff
@@ -194,6 +194,7 @@ impl Config for Runtime {
     type Oracle = oracle_proxy::Pallet<Runtime>;
     type Symbol = <Runtime as band::Config>::Symbol;
     type GetSyntheticBaseBuySellLimit = GetSyntheticBaseBuySellLimit;
+    type TradingPairSourceManager = trading_pair::Pallet<Runtime>;
```
wer1st, line 197: You need to add it into runtime as well

wer1st (approved):

PR #707, opened 2023-09-06
Improve batch swap transfer

vovac12:


pallets/liquidity-proxy/src/lib.rs
```diff
@@ -1964,24 +1973,32 @@ impl<T: Config> Pallet<T> {
                     fail!(Error::<T>::InvalidReceiversInfo);
                 }
 
-                let out_amount = receivers.iter().map(|recv| recv.target_amount).sum();
+                let out_amount = receivers
+                    .iter()
+                    .map(|recv| recv.target_amount)
+                    .sum::<Balance>()
```
wer1st, line 1979: `sum()` can overflow and panic

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "nightly-2023-03-21"
+targets = [ "wasm32-unknown-unknown" ]
+profile = "default"
```
wer1st, line 4: add empty line please

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
```
Alexey-N-Chernyshov, line 1: As I remember, there was a reason not to commit toolchain file. I could be wrong.

wer1st (approved):

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
...
+channel = "nightly-2023-03-21"
```
Alexey-N-Chernyshov, line 2: When we decide to change toolchain, we must update this line and [this line](https://github.com/sora-xor/sora2-network/blob/master/housekeeping/docker/env/Dockerfile#L47).
Probably we want to do it now for #709

rust-toolchain.toml
Alexey-N-Chernyshov, line 4:
```suggestion
profile = "default"

```

Alexey-N-Chernyshov (approved):

rust-toolchain.toml
vovac12: Actually it doesn't build without it. And I think we can use it to ensure that correct Rust version is used by everyone. If we will have any problems with this solution we can remove this file

rust-toolchain.toml
vovac12: We have the same version of Rust in docker and we can't update it before Substrate update because of problems with wasm

Alexey-N-Chernyshov (approved):

PR #709, opened 2023-09-06
Tests for payout_stakers inside batch

pakrentos:


Alexey-N-Chernyshov (approved):
rust version required is 1.72, actual is 1.68

PR #717, opened 2023-09-11
Benchmark results of benchmarking-6 from develop

sorabot:
Check the benchmarking results.
Results for special cases we would like to review.


bragov4ik:
did not enable all benchmarks 😮‍💨😮‍💨😮‍💨

PR #732, opened 2023-09-19
ALT: New smart split of liquidity

wer1st:


pallets/dex-api/src/lib.rs
```diff
@@ -122,6 +123,41 @@ impl<T: Config>
...
+            XSTPool => step_quote!(XSTPool),
+
+            #[cfg(feature = "wip")] // order-book
+            OrderBook => step_quote!(OrderBook),
+
+            MockPool => step_quote!(MockLiquiditySource),
+            MockPool2 => step_quote!(MockLiquiditySource2),
+            MockPool3 => step_quote!(MockLiquiditySource3),
+            MockPool4 => step_quote!(MockLiquiditySource4),
+            BondingCurvePool => unreachable!(),
```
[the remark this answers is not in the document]
wer1st: Probably for some mocks

pallets/multicollateral-bonding-curve-pool/src/lib.rs
```diff
@@ -1588,6 +1589,59 @@ impl<T: Config> LiquiditySource<T::DEXId, T::AccountId, T::AssetId, Balance, Dis
...
+        let mut chunks = VecDeque::new();
+        let mut sub_in = Balance::zero();
+        let mut sub_out = Balance::zero();
+        let mut sub_fee = Balance::zero();
+
+        for i in 1..=samples_count {
+            let volume = amount.copy_direction(
+                step.checked_mul(i as Balance)
+                    .ok_or(Error::<T>::ArithmeticError)?,
+            );
```
bragov4ik, line 1623: If I'm not mistaken, when `i = samples_count`, `volume` might be smaller than `amount` (because of rounding on `checked_div`), so the sum of chunks is smaller than requested `amount`. Not sure about use cases of `step_quote` and whether this might be a problem.

pallets/order-book/src/lib.rs
```diff
@@ -1086,6 +1087,70 @@ impl<T: Config> LiquiditySource<T::DEXId, T::AccountId, T::AssetId, Balance, Dis
         }
     }
 
+    fn step_quote(
+        dex_id: &T::DEXId,
+        input_asset_id: &T::AssetId,
+        output_asset_id: &T::AssetId,
+        amount: QuoteAmount<Balance>,
+        _samples_count: usize,
```
bragov4ik, line 1095: Should this parameter be even included in the interface, if one of the implementations ignores it? Sounds tricky to use with implicit/unexpected behaviour

pallets/order-book/src/order_book.rs
```diff
@@ -1167,4 +1168,59 @@ impl<T: crate::Config + Sized> OrderBook<T> {
...
+                data.get_aggregated_bids(&self.order_book_id).iter().rev(),
+            ),
+            PriceVariant::Sell => Self::get_market_depth(
+                volume_limit,
+                data.get_aggregated_asks(&self.order_book_id).iter(),
+            ),
+        }
+    }
+
+    fn get_market_depth<'a>(
```
bragov4ik, line 1190: minor thing, but imho something like `stripped_market_data` or `limited_market_data` or `bounded_market_data` would be more precise for the implementation

pallets/xst/src/lib.rs
```diff
@@ -1087,6 +1090,58 @@ impl<T: Config> LiquiditySource<T::DEXId, T::AccountId, T::AssetId, Balance, Dis
...
+            fail!(Error::<T>::CantExchange);
+        }
+        if amount.amount().is_zero() {
+            return Ok(VecDeque::new());
+        }
+
+        let synthetic_base_asset_id = &T::GetSyntheticBaseAssetId::get();
+        let (input_amount, output_amount, fee_amount) = if input_asset_id == synthetic_base_asset_id
+        {
+            Self::decide_sell_amounts(&input_asset_id, &output_asset_id, amount, deduce_fee, false)?
```
bragov4ik, line 1111: In `quote`, `check_limits` is set to `true`. Why `false` here?

Did not dig into implementation of `xst` pallet; would be glad if it was clarified here (maybe as comments?)

pallets/xst/src/lib.rs
```diff
@@ -1087,6 +1090,58 @@ impl<T: Config> LiquiditySource<T::DEXId, T::AccountId, T::AssetId, Balance, Dis
...
+
+        let ratio = (FixedWrapper::from(1) / FixedWrapper::from(samples_count))
+            .get()
+            .map_err(|_| Error::<T>::PriceCalculationFailed)?;
+
+        let chunk = monolith
+            .rescale_by_ratio(ratio)
+            .ok_or(Error::<T>::PriceCalculationFailed)?;
+
+        Ok(vec![chunk; samples_count].into())
```
bragov4ik, line 1142: isn't it possible to use `inner_quote` since no splitting into chunks occurs?

bragov4ik: maybe I miss something, did not check thoroughly

pallets/liquidity-proxy/src/lib.rs
```diff
@@ -1412,6 +1382,48 @@ impl<T: Config> Pallet<T> {
         }
     }
 
+    fn new_smart_split(
```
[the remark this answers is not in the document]
wer1st: when it goes to prod, it will be called `smart_split`. Old smart split function will be deleted

pallets/liquidity-proxy/src/lib.rs
```diff
@@ -1412,6 +1382,48 @@ impl<T: Config> Pallet<T> {
...
+    ) -> Result<
+        (
+            AggregatedSwapOutcome<LiquiditySourceIdOf<T>, Balance>,
+            Rewards<T::AssetId>,
+            Weight,
+        ),
+        DispatchError,
+    > {
+        ensure!(
+            input_asset_id == base_asset_id || output_asset_id == base_asset_id,
```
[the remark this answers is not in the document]
wer1st: It is taken from old `smart_split`.
It is necessary to check that XOR is a part of swap. If user wants to exchange Asset1 to Asset2 - it must be splitter into 2 swaps: Asset1 -> XOR -> Asset2.
If `smart_split` receives the assets without XOR - it must be denied.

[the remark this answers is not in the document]
wer1st: And the same for XSTUSD for dex_id = 1

pallets/liquidity-proxy/src/liquidity_aggregator.rs
```diff
@@ -0,0 +1,1123 @@
...
+            fee,
+        }
+    }
+}
+
+/// Aggregates the liquidity from the provided liquidity sources.
+/// Liquidity sources provide discretized liquidity curve by chunks and then Liquidity Aggregator selects the best chunks from different sources to gain the best swap amount.
+#[derive(Clone)]
+pub struct LiquidityAggregator<LiquiditySourceType> {
+    liquidity_chunks: BTreeMap<LiquiditySourceType, VecDeque<SwapChunk<Balance>>>,
```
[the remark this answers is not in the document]
wer1st: It require to pre-calculate all prices for all chunks. In the current solution we calculate price only for front chunks.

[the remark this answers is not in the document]
wer1st: also BTreeMap sorts it from low to high

pallets/multicollateral-bonding-curve-pool/src/lib.rs
wer1st: Yes, it is acceptable rounding that will be covered by slippage

pallets/order-book/src/lib.rs
wer1st: But we need to provide it to another sources. It is ok if some implementations don't use all attribites from time to time

pallets/xst/src/lib.rs
wer1st: it is split, it returns vector of equal chunks

pallets/liquidity-proxy/src/lib.rs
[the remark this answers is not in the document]
bragov4ik: Oh, I see, thanks.

pallets/liquidity-proxy/src/liquidity_aggregator.rs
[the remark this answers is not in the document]
bragov4ik: > It require to pre-calculate all prices for all chunks. In the current solution we calculate price only for front chunks.

So it assumes all `VecDeque<SwapChunk<Balance>>`s that are provided via `add_source` are pre-sorted by price? Is my understanding correct?

If that's the case, I suggest to place some safety net, like checking that the provided `chunks` is sorted or at least specifying it in the function signature/comments:
```rust
pub fn add_source(
    ...,
    sorted_chunks: VecDeque<SwapChunk<Balance>>,
)
```

pallets/order-book/src/lib.rs
bragov4ik: Hmm, maybe call it something like `samples_count_default` so that the intent is more clear when using the trait?

pallets/xst/src/lib.rs
bragov4ik: Got it, thanks!
I think adding short explanation in the nearby comments would help future developers; it doesn't seem immediately obvious

pallets/xst/src/lib.rs
bragov4ik: Missed it, ok)

pallets/liquidity-proxy/src/liquidity_aggregator.rs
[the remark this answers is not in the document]
wer1st: Yes, they are sorted by price, but not manually. It is a specific of liquidity curves that each next chunk is worse or the same.

pallets/order-book/src/lib.rs
wer1st: `recommended_samples_count`

pallets/xst/src/lib.rs
bragov4ik: thanks for adding it 👍

bragov4ik (changes requested):
The only bit that bothers me is zero weight, going to approve after discussing/fixing it

pallets/multicollateral-bonding-curve-pool/src/lib.rs
```diff
@@ -1594,29 +1594,29 @@ impl<T: Config> LiquiditySource<T::DEXId, T::AccountId, T::AssetId, Balance, Dis
...
+        recommended_samples_count: usize,
         deduce_fee: bool,
-    ) -> Result<VecDeque<SwapChunk<Balance>>, DispatchError> {
+    ) -> Result<(VecDeque<SwapChunk<Balance>>, Weight), DispatchError> {
         if !Self::can_exchange(dex_id, input_asset_id, output_asset_id) {
             fail!(Error::<T>::CantExchange);
         }
         if amount.amount().is_zero() {
-            return Ok(VecDeque::new());
+            return Ok((VecDeque::new(), Weight::zero()));
```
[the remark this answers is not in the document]
wer1st: zero is okay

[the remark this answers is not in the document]
bragov4ik: Ok, I guess it shouldn't be a big problem, but my point was that calculations done before this return should be considered in weight.

It just seems for me as an easy approach for weight evaluation (particularly, since it's quite [important](https://docs.substrate.io/test/benchmark/#why-benchmark-a-pallet) to do so). I think it's easier to consider all computations that happen in final weight rather than checking case-by-case whether zero weight here can lead to some attacks or not (although in this the answer is "likely not", so I'm going to approve)

bragov4ik (approved):

pallets/liquidity-proxy/src/lib.rs
```diff
@@ -954,11 +933,6 @@ impl<T: Config> Pallet<T> {
             T::LiquidityRegistry::list_liquidity_sources(input_asset_id, output_asset_id, filter)?;
         let locked = T::LockedLiquiditySourcesManager::get();
         sources.retain(|x| !locked.contains(&x.liquidity_source_index));
-        // The temp solution is to exclude OrderBook source if there are multiple sources.
-        // Will be redesigned in #447
-        if sources.len() > 1 {
```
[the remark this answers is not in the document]
wer1st: Yes, it is better to keep it under the feature flag for the old mechanism

vovac12 (approved):

PR #734, opened 2023-09-20
Order book attributes benchmark

bragov4ik:
This PR contains benchmarks to figure out appropriate order book parameters.
In particular, extrinsics of `order-book` pallet must always fit into one block (and not delay block production)

Main contributions:
* Worst-case benchmarks for heavy extrinsics/public interface in order book are created. They are configurable by order book parameters.
* The benchmarks are tested (with manually-copied code for now): the easiest way to test the benchmarks without waiting for slow compilation.
* Benchmarking different parameters sets:
  * The parameter sets (presets) are configured manually (in compile time) in `preparation.rs`.
  * Python script that generates benchmarks for these presets (macros did not work because benchmarks are already in a substrate's `benchmarks!` macro). The benchmarks are marked as `#[extra]` since they are needed only for development.
  * Script for running the benchmarks automatically is generated in the same script.
* Split `benchmarking.rs` into the benchmarks themselves and code for preparing them.

Other (collateral) changes:
* Fixed aggregated price storage length check. It now considers addition of order into existing price entry.
* Storage. Changes were made to avoid cloning when not necessary:
  * Mutable reference to cached items of a map/double map.
  * Determine fullness of bounded item in a storage (e.g. `BoundedVec`) without retrieving the whole item (in a similar way to `StorageDecodeLength`).
  * `CachedDataLayer` is optimized by utilizing the aforementioned changes.
  * `StorageDataLayer` is optimized by utilizing the aforementioned changes and `mutate`. 
* Explicit multiplication of `BalanceUnit` by scalar (`Scalar` newtype): *in order to clearly see the intention of a programmer; multiplying by just divisible `BalanceUnit` might be misleading*.
* Pretty printing of order book contents and expiration schedule. Should be helpful for local (test) debugging.
* `test_utils` are made available for benchmarks as well; organized into modules.


pallets/order-book/src/traits.rs
```diff
@@ -104,25 +104,58 @@ where
         price: &OrderPrice,
     ) -> Option<PriceOrders<T::OrderId, T::MaxLimitOrdersForPrice>>;
 
+    /// Returns whether there is no place for orders inside the bid price
+    fn is_bids_full(
```
wer1st, line 108:
```suggestion
    fn are_bids_full(
```

pallets/order-book/src/traits.rs
```diff
@@ -104,25 +104,58 @@ where
...
+    ) -> Option<bool>;
+
     /// Returns order ids of orders inside the ask price
     fn get_asks(
         &mut self,
         order_book_id: &OrderBookId<AssetIdOf<T>, T::DEXId>,
         price: &OrderPrice,
     ) -> Option<PriceOrders<T::OrderId, T::MaxLimitOrdersForPrice>>;
 
+    fn is_asks_full(
```
wer1st, line 121:
```suggestion
    fn are_asks_full(
```

pallets/order-book/src/traits.rs
```diff
@@ -138,6 +171,13 @@ where
         OrderBookId<AssetIdOf<T>, T::DEXId>,
         UserOrders<T::OrderId, T::MaxOpenedLimitOrdersPerUser>,
     >;
+
+    /// Returns whether there is no place for the user's orders in the order book
+    fn is_user_limit_orders_full(
```
wer1st, line 176:
```suggestion
    fn are_user_limit_orders_full(
```

pallets/order-book/src/cache_data_layer.rs
```diff
@@ -439,6 +460,16 @@ impl<T: Config> DataLayer<T> for CacheDataLayer<T> {
         self.asks.get(order_book_id, price).cloned()
     }
 
+    fn is_asks_full(
```
wer1st, line 463: you need to add tests for all new methods in `tests/data_layer.rs`

wer1st: the same for `storage_data_layer.rs`

pallets/order-book/src/order_book.rs
```diff
@@ -416,7 +416,7 @@ impl<T: crate::Config + Sized> OrderBook<T> {
             deal_input: None,
             deal_output: None,
             market_input: None,
-            market_output: None, // NA for this case, because all the liquidity of both types go out of market
+            market_output: None, // NA for this case, because all the liquidity of both types goes out of market
```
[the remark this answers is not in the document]
bragov4ik: The correction is correct (?): liquidity is uncountable

[_Uncountable nouns refer to things that we cannot count. Such nouns take only singular form._](https://owl.purdue.edu/owl/general_writing/grammar/count_and_noncount_nouns/count_noncount_nouns_with_articles_adjectives.html#:~:text=Uncountable%20nouns%20refer%20to%20things,nouns%20take%20only%20singular%20form.)

common/src/storage.rs
```diff
@@ -0,0 +1,232 @@
+use codec::{EncodeLike, FullCodec, FullEncode};
```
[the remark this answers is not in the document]
bragov4ik: added, thx

pallets/order-book/src/traits.rs
bragov4ik: My logic was that `bids` is a storage value, so this checks if the value is full.
It seems non-intuitive, I agree. What do you think about `is_bid_price_full`?

pallets/order-book/src/traits.rs
bragov4ik: Same logic here, "is storage value of `user_limit_orders` full?" 
Not sure what would mean `are_user_limit_orders_full`, seems to imply that orders themselves are full 🤔

pallets/order-book/src/traits.rs
wer1st: `is_bid_price_full` looks incorrect

pallets/order-book/src/cache_data_layer.rs
bragov4ik: done

bragov4ik: btw tests for length exposed a nasty bug, much thanks :)

pallets/order-book/src/traits.rs
wer1st: Ok

pallets/order-book/src/benchmarking/mod.rs
```diff
@@ -0,0 +1,1986 @@
...
+        }: {
+            OrderBookPallet::<T>::execute_market_order(
+                RawOrigin::Signed(caller.clone()).into(),
+                order_book_id,
+                PriceVariant::Sell,
+                *amount.balance()
+            ).unwrap();
+        }
+        verify {
+            let average_price = BalanceUnit::divisible(325000000000000);
```
wer1st, line 476: Please use `balance!()` in all places

runtime/src/lib.rs
```diff
@@ -1913,10 +1913,19 @@ impl order_book::Config for Runtime {
...
+    // to allocate (next_power_of_two) bytes anyway.
+    //
+    // or initialize it via `with_capacity` instead.
+    //
+    // this limit is mostly because of requirement to use bounded vectors.
+    // a user can create multiple accounts at any time.
+    type MaxOpenedLimitOrdersPerUser = ConstU32<1024>;
+    type MaxLimitOrdersForPrice = ConstU32<64>; // TODO #711 #723: try to increase the value
+    type MaxSidePriceCount = ConstU32<64>; // TODO #711 #723: try to increase the value
+    type MaxExpiringOrdersPerBlock = ConstU32<512>; // TODO #722: clarify parameter
```
wer1st, line 1928: Actualize it please

wer1st (approved):

wer1st:
The remaining default values of the attributes:
`MAX_ORDER_LIFESPAN` - 1 month
`MIN_ORDER_LIFESPAN` - 1 min
`MAX_PRICE_SHIFT` - 50 %
`tick_size` - 0.00001
`step_lot_size` - 0.00001
`min_lot_size` - 1
`max_lot_size` - 1000

wer1st:
@bragov4ik , the remaining default values of the attributes:
`MAX_ORDER_LIFESPAN` - 1 month
`MIN_ORDER_LIFESPAN` - 1 min
`MAX_PRICE_SHIFT` - 50 %
`tick_size` - 0.00001
`step_lot_size` - 0.00001
`min_lot_size` - 1
`max_lot_size` - 1000

Approved by @Tieumsan

bragov4ik:
> `tick_size` - 0.00001

@wer1st Wouldn't this combined with _`MaxSidePriceCount`  - 1024_ make it easy to clog the order book with 1024 orders with some unrealistic prices?

For example, placing a buy order for each price 0.00001...0.01024 while a realistic marketed price is ~10-100 would not allow placing orders outside the range. Of course it considers that the order book is completely empty at the start, so I'm not sure how problematic this can become.

bragov4ik:
Also, @wer1st, regarding indivisible assets. I suppose the parameters would be

tick_size - same
step_lot_size - 1 (indivisible)
min_lot_size - 1 (indivisible)
max_lot_size - 1000 (indivisible)

Is this correct?

wer1st:
> > `tick_size` - 0.00001
> 
> @wer1st Wouldn't this combined with _`MaxSidePriceCount` - 1024_ make it easy to clog the order book with 1024 orders with some unrealistic prices?
> 
> For example, placing a buy order for each price 0.00001...0.01024 while a realistic marketed price is ~10-100 would not allow placing orders outside the range. Of course it considers that the order book is completely empty at the start, so I'm not sure how problematic this can become.

We discussed this case, it is ok. Keep this values as default.

wer1st:
> Also, @wer1st, regarding indivisible assets. I suppose the parameters would be
> 
> tick_size - same step_lot_size - 1 (indivisible) min_lot_size - 1 (indivisible) max_lot_size - 1000 (indivisible)
> 
> Is this correct?

Yes

pallets/order-book/src/lib.rs
```diff
@@ -32,6 +32,9 @@
 #![allow(dead_code)] // todo (m.tagirov) remove
 // TODO #167: fix clippy warnings
 #![allow(clippy::all)]
+// too many benchmarks, doesn't compile otherwise
+#![recursion_limit = "512"]
+#![feature(int_roundings)]
```
wer1st, line 37: What if we move it in `limit_order.rs`, because I don't see how it is used in the current file

wer1st: Is it available with nightly only?

pallets/order-book/src/test_utils.rs
```diff
@@ -0,0 +1,920 @@
...
+
+#[allow(unused)]
+#[cfg(not(test))]
+use crate::{
+    self as order_book, cache_data_layer::CacheDataLayer, traits::DataLayer, Asks, Bids, Config,
+    Event, ExpirationScheduler, ExpirationsAgenda, LimitOrder, LimitOrders, MarketRole, MomentOf,
+    OrderAmount, OrderBook, OrderBookId, OrderBookStatus, OrderBooks, OrderPrice, OrderVolume,
+    Pallet, Payment, PriceOrders,
+};
+#[allow(unused)]
```
wer1st, line 41: Why do we need it? Do we have some unused methods?

pallets/order-book/src/lib.rs
bragov4ik: 1. It's a crate attribute, so it must be placed in the crate root file
2. It was stabilized yesterday in Rust 1.73. However, we use nightly (2023-03-21?), thus I'm enabling the feature. Btw it seems to work on stable 1.73+ as well (by ignoring the attribute, I suppose).

pallets/order-book/src/test_utils.rs
bragov4ik: fixed it, also removed duplications of the imports

wer1st (approved):

pallets/order-book/src/benchmarking/mod.rs
```diff
@@ -0,0 +1,1942 @@
...
+            caller_quote_balance + expected_out
+        );
+    }
+}
+
+#[cfg(not(test))]
+pub use benchmarks_inner::*;
+
+#[cfg(not(test))]
+mod benchmarks_inner {
```
vovac12, line 539: This module is big enough, so I think you can move it into separate file. Also order book benchmarks now is big and complex, so I think we can also move it to separate crate as it's done for some another pallets

vovac12 (commented):
Could you please add weights computed locally as initial reference? It will help to review Config values properly

pallets/order-book/src/benchmarking/mod.rs
bragov4ik: I extracted the benchmarks into a separate crate. Whilst doing this the `lib.rs` file basically consists of `benchmarks_inner` module, so I left it there. Just to clarify, the module is needed to not compile the `benchmarks!` macro for tests; otherwise I'd put it without the module.

However, I couldn't manage to fix wasm compilation for `--tests` :( without tests it works, as well as with tests and not wasm. can't figure out the problem since I basically copied .toml from order-book :/
part of error message for context
```
error[E0152]: duplicate lang item in crate `std` (which `order_book_benchmarking` depends on): `panic_impl`
```

not sure how to fix though; maybe try to copy some benchmarking crate's toml?

wer1st: Fixed

bragov4ik: Also ran the benchmarks locally for the reference

wer1st (approved):

bragov4ik:
removed fix for #789 because solved more elegantly in #795

pallets/order-book/src/weights.rs
```diff
@@ -48,6 +74,14 @@ pub trait WeightInfo {
         Weight::zero()
     }
     fn cancel_limit_order() -> Weight {
+        // The functions for first/last expiring order are generated automatically.
+        // This implementation is not produced by benchmarking.
+        Self::cancel_limit_order_first_expiration().max(Self::cancel_limit_order_last_expiration())
```
vovac12, line 79: weights.rs should not be modified after generation to make easier automatization of benchmarking

pallets/order-book/src/weights.rs
```diff
@@ -74,42 +108,289 @@ impl WeightInfo for () {}
...
+    /// Proof: System Account (max_values: None, max_size: Some(128), added: 2603, mode: MaxEncodedLen)
+    /// Storage: OrderBook ExpirationsAgenda (r:8 w:8)
+    /// Proof: OrderBook ExpirationsAgenda (max_values: None, max_size: Some(43014), added: 45489, mode: MaxEncodedLen)
     fn place_limit_order() -> Weight {
-        Weight::zero()
+        // Proof Size summary in bytes:
+        //  Measured:  `1313466`
+        //  Estimated: `14128490`
+        // Minimum execution time: 170_342_000 nanoseconds.
+        Weight::from_parts(171_938_000_000, 14128490)
```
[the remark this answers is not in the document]
bragov4ik: Fixed, now it measures only limit order placement without conversion to market order(-s).

pallets/order-book/src/weights.rs
bragov4ik: Now it's unaltered after generation.

wer1st (approved):

vovac12 (approved):
LGTM

PR #755, opened 2023-10-02
OrderBook: exchange dynamic weight

wer1st:


runtime/src/lib.rs
```diff
@@ -1907,6 +1907,8 @@ impl order_book::Config for Runtime {
     const MIN_ORDER_LIFESPAN: Moment = MILLISECS_PER_BLOCK; // TODO: order-book clarify
     const MILLISECS_PER_BLOCK: Moment = MILLISECS_PER_BLOCK;
     const MAX_PRICE_SHIFT: Perbill = Perbill::from_percent(50); // TODO: order-book clarify
+    const SMALL_MIN_MAX_RATIO: usize = 1000;
+    const BIG_MIN_MAX_RATIO: usize = 4000;
```
bragov4ik, line 1911: I'm thinking if instead of SMALL/BIG naming it SOFT/HARD would make more sense 🤔  (as in soft/hard limit)

pallets/order-book/src/lib.rs
```diff
@@ -677,12 +679,27 @@ pub mod pallet {
...
 
-            Ok(().into())
+            // Note: be careful with changing the weight. The fee depends on it,
+            // the market-maker fee is charged for some weight, and the regular fee for none weight
+            let actual_weight = if executed_orders_count == 0 {
+                // if the extrinsic just places the limit order, the weight of the placing is returned
+                Some(<T as Config>::WeightInfo::place_limit_order())
+            } else {
+                // if the limit order was converted into market order, then None weight is returned
+                // this weight will be replaced with worst case weight - exchange_weight()
```
bragov4ik, line 695: Is this logic planned to be implemented in other PR? I don't see it here

bragov4ik: Also didn't we want to charge fee proportional to the executed number instead?

bragov4ik (changes requested):

pallets/order-book/src/lib.rs
wer1st: Not propotional, just worst case.
It is already implemented, look at `#[pallet::weight(Pallet::<T>::exchange_weight())]`. This weight will be used if extrinsic returns `None` for the weight.

bragov4ik: i though about another piece of logic (fee calc); it was clarified to me that it's going to be resolved in a following PR

pallets/order-book/src/lib.rs
```diff
@@ -114,6 +114,8 @@ pub mod pallet {
         const MIN_ORDER_LIFESPAN: MomentOf<Self>;
         const MILLISECS_PER_BLOCK: MomentOf<Self>;
         const MAX_PRICE_SHIFT: Perbill;
+        const SMALL_MIN_MAX_RATIO: usize;
```
vovac12, line 117: It would be nice to have a docs for this parameters

vovac12 (approved):

PR #772, opened 2023-10-09
Updating order book attributes optimization

wer1st:


pallets/order-book/src/tests/test_utils.rs
```diff
@@ -328,3 +328,20 @@ pub fn run_to_block(n: u32) -> Weight {
...
+pub fn update_orderbook_unchecked(
+    order_book_id: OrderBookId<AssetIdOf<Runtime>, DEXId>,
+    tick_size: Balance,
+    step_lot_size: Balance,
+    min_lot_size: Balance,
+    max_lot_size: Balance,
+) -> OrderBook<Runtime> {
+    let mut order_book = OrderBookPallet::order_books(order_book_id).unwrap();
+    order_book.tick_size.set(tick_size);
+    order_book.step_lot_size.set(step_lot_size);
```
bragov4ik, line 341: this doesn't perform any alignments, correct?

bragov4ik: just want to make sure it's intended

pallets/order-book/src/lib.rs
```diff
@@ -169,12 +168,7 @@ pub mod pallet {
         type SyntheticInfoProvider: SyntheticInfoProvider<Self::AssetId>;
         type DexInfoProvider: DexInfoProvider<Self::DEXId, DEXInfo<Self::AssetId>>;
         type Time: Time;
-        type ParameterUpdateOrigin: EnsureOrigin<
-            Self::RuntimeOrigin,
-            Success = Either<Self::AccountId, ()>,
-        >;
-        type StatusUpdateOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = ()>;
-        type RemovalOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = ()>;
+        type PermissionOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = ()>;
```
bragov4ik, line 171: It seems something like `PermittedOrigin` would be more accurate. I might be misunderstanding the intention here; What do you think?

bragov4ik (approved):
LGTM

pallets/order-book/src/tests/test_utils.rs
wer1st: Yes, without alignment. Just set raw values without any checks. It's necessary to make the preconditions setup more easy for some tests.

vovac12 (approved):
LGTM

PR #793, opened 2023-10-21
Refactor/mv relayer

Alexey-N-Chernyshov:
Moved to https://github.com/sora-xor/bridge-relayer.


substrate-gen/src/main.rs
vovac12, line 1: Why not make this file `build.rs`?

Alexey-N-Chernyshov: Why? We don't need it in every build. When it is time to generate the file, we can run it with command `cargo run -p substrate-gen`.

wer1st (approved):

vovac12 (approved):

PR #801, opened 2023-10-24
Trading pair source trait

TheLedgerOfJoudi:
Added the TradingPairSourceManager trait to pallet configs in order to avoid direct calls from the trading-pair pallet.

1. All components get enabled sources info from TradingPairSourceManager : Yes.
2. All tests passed: Yes.


node/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "framenode"
-version = "3.0.0"
+version = "3.1.0"
```
wer1st, line 3: You don't need to change the version of any pallet in this scope

runtime/src/lib.rs
```diff
@@ -256,10 +256,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("sora-substrate"),
     impl_name: create_runtime_str!("sora-substrate"),
     authoring_version: 1,
-    spec_version: 64,
+    spec_version: 66,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
-    transaction_version: 64,
+    transaction_version: 66,
```
wer1st, line 262: Please revert it. We have a special process where we update these values

TheLedgerOfJoudi:
I moved this draft to new PR #808, kindly check. Pumping versions appears here because I did a wrong branch-out before opening the PR.

PR #803, opened 2023-10-26
OrderBook: update events

wer1st:


pallets/order-book/src/lib.rs
```diff
@@ -346,14 +350,23 @@ pub mod pallet {
             order_id: T::OrderId,
             owner_id: T::AccountId,
             side: PriceVariant,
+            price: OrderPrice,
             amount: OrderAmount,
         },
 
+        /// All amount of the limit order is executed
+        LimitOrderFilled {
```
bragov4ik, line 358: `LimitOrderFulfilled` seems more accurate/readable, do you agree?

pallets/order-book/src/types.rs
```diff
@@ -67,6 +67,17 @@ pub enum OrderBookStatus {
     Stop,
 }
 
+#[derive(
+    Encode, Decode, PartialEq, Eq, Copy, Clone, Debug, scale_info::TypeInfo, MaxEncodedLen,
+)]
+pub enum CancelReason {
+    /// User cancels the limit order by themselves
```
bragov4ik, line 74:
```suggestion
    /// User cancels the limit order by themself
```
minor correction 🙃

bragov4ik (approved):
LGTM, but I want to clarify:
there's a point in the task that says
- `LimitOrderIsSplitIntoMarketOrderAndLimitOrder` should go together with `LimitOrderPlaced` 

Is it addressed here? I just didn't quite catch that

pallets/order-book/src/lib.rs
wer1st: Not. `Filled` is a common status for different exchanges

pallets/order-book/src/types.rs
wer1st: Both variant are ok, because of moder notation. But ok

wer1st:
> LGTM, but I want to clarify: there's a point in the task that says
> 
> * `LimitOrderIsSplitIntoMarketOrderAndLimitOrder` should go together with `LimitOrderPlaced`
> 
> Is it addressed here? I just didn't quite catch that

It already works

vovac12 (approved):

PR #806, opened 2023-10-27
Benchmark results of benchmarking-order-book-fix-2 from develop

sorabot:
Check the benchmarking results


bragov4ik:
results are quite unexpected 😮

PR #814, opened 2023-11-03
Fix XST pool buy/sell limit

vovac12:


Alexey-N-Chernyshov (approved):

wer1st (approved):

Nolan730:
Can we get clarification about this? It seems this is going to limit all syntetic assets with 10m xst. Also suprised there isnt 18 zero after it but i doubt its needed since you cant make this mistake?

vovac12:
> Can we get clarification about this? It seems this is going to limit all syntetic asset swaps with 10m xst. Also suprised there isnt 18 zero after it but i doubt its needed since you cant make this mistake?

`balance!` macro adds 18 zero. You can check this value on the stage environment via polkadot.js.org/apps in Developer > Chain state > Constants > xstPool > getSyntheticBaseBuySellLimit

PR #816, opened 2023-11-07
QA Tools improvements for order book

bragov4ik:
- More flexible settings for order book create&fill in `qa-tools`
- Some refactoring of `qa-tools`


common/src/balance_unit.rs
```diff
@@ -179,6 +180,29 @@ impl BalanceUnit {
         };
         Ok(Self::new(balance, self.is_divisible))
     }
+
+    pub fn into_indivisible(mut self, mode: RoundMode) -> Self {
```
bragov4ik, line 184: this function is not needed after changing the `fill_order_book`. I think I'll remove it soon, since there's no demand

pallets/qa-tools/src/pallet_tools/order_book.rs
```diff
@@ -1,26 +1,72 @@
...
+        /// Best price = highest, worst = lowest.
+        pub bids: SideFill,
+        /// Best price = lowest, worst = highest.
+        pub asks: SideFill,
+        /// Lifespan of inserted orders, max by default.
+        pub lifespan: Option<Moment>,
+        /// Seed for producing random values during the fill process. If `None`,
+        /// current block is chosen
+        pub random_seed: Option<BlockNumber>,
+    }
```
bragov4ik, line 63: @ra9mls can you check preliminary that such settings are ok? (they are to be set for each order book id in the extrinsic)

pallets/qa-tools/src/pallet_tools/liquidity_proxy.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub struct XYKPair<DEXId, AssetId> {
+        pub dex_id: DEXId,
+        pub asset_a: AssetId,
+        pub asset_b: AssetId,
+        /// Price of `asset_a` in terms of `asset_b` (how much `asset_b` is needed to buy 1 `asset_a`)
+        pub price: BalanceUnit,
+    }
+
+    #[allow(unused)]
+    pub fn xst() {}
```
wer1st, line 22: please remove

pallets/qa-tools/src/pallet_tools/liquidity_proxy.rs
```diff
@@ -0,0 +1,56 @@
...
+    use sp_std::vec::Vec;
+
+    #[derive(Clone, PartialEq, Eq, Encode, Decode, scale_info::TypeInfo, Debug)]
+    #[scale_info(skip_type_params(T))]
+    pub struct XYKPair<DEXId, AssetId> {
+        pub dex_id: DEXId,
+        pub asset_a: AssetId,
+        pub asset_b: AssetId,
+        /// Price of `asset_a` in terms of `asset_b` (how much `asset_b` is needed to buy 1 `asset_a`)
+        pub price: BalanceUnit,
```
wer1st, line 18: Probably we can use just `Balance` type here, because XYK pool works only with divisible assets.

pallets/qa-tools/src/pallet_tools/liquidity_proxy.rs
```diff
@@ -0,0 +1,56 @@
...
+    use common::prelude::BalanceUnit;
+    use frame_support::dispatch::DispatchResult;
+    use frame_system::pallet_prelude::BlockNumberFor;
+    use order_book::{MomentOf, OrderBookId};
+    use sp_std::fmt::Debug;
+    use sp_std::vec::Vec;
+
+    #[derive(Clone, PartialEq, Eq, Encode, Decode, scale_info::TypeInfo, Debug)]
+    #[scale_info(skip_type_params(T))]
+    pub struct XYKPair<DEXId, AssetId> {
```
wer1st, line 13: Where is it used?

common/src/lib.rs
```diff
@@ -31,6 +31,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 // TODO #167: fix clippy warnings
 #![allow(clippy::all)]
+#![feature(int_roundings)]
```
wer1st, line 34: For what?

pallets/qa-tools/src/pallet_tools/liquidity_proxy.rs
bragov4ik: This was adding when working on another qa-tools task, so I accidentally brought it here :)

common/src/lib.rs
bragov4ik: no reason, after I've deleted the now-redundant functions

stefashkaa (approved):

wer1st (approved):

pallets/order-book/src/lib.rs
```diff
@@ -692,7 +692,7 @@ pub mod pallet {
                 // schedule alignment
                 <AlignmentCursor<T>>::set(order_book_id, Some(T::OrderId::zero()));
             }
-            <OrderBooks<T>>::set(order_book_id, Some(order_book));
+            <OrderBooks<T>>::insert(order_book_id, order_book);
```
wer1st, line 695: I prefer to keep `set`

pallets/qa-tools/src/lib.rs
```diff
@@ -82,6 +84,13 @@ pub mod pallet {
     pub type WhitelistedCallers<T: Config> =
         StorageValue<_, BoundedBTreeSet<T::AccountId, T::QaToolsWhitelistCapacity>>;
 
+    #[pallet::event]
+    #[pallet::generate_deposit(pub(super) fn deposit_event)]
+    pub enum Event<T: Config> {
+        AddedToWhitelist { user: T::AccountId },
+        RemovedFromWhitelist { user: T::AccountId },
+    }
+
```
wer1st, line 93: You don't need to introduce events because I will delete the whitelist in my PR

pallets/qa-tools/Cargo.toml
```diff
@@ -16,12 +16,14 @@ codec = { package = "parity-scale-codec", version = "3", default-features = fals
...
 common = { path = "../../common", default-features = false }
+rand = { version = "0.8.5", default-features = false }
+rand_chacha = { version = "0.3.1", default-features = false }
 scale-info = { version = "2", default-features = false, features = ["derive"] }
 sp-std = { git = "https://github.com/sora-xor/substrate.git", branch = "polkadot-v0.9.38", default-features = false }
 frame-support = { git = "https://github.com/sora-xor/substrate.git", branch = "polkadot-v0.9.38", default-features = false }
 frame-system = { git = "https://github.com/sora-xor/substrate.git", branch = "polkadot-v0.9.38", default-features = false }
 assets = { path = "../assets", default-features = false }
-order-book = { path = "../../pallets/order-book", default-features = false }
+order-book = { path = "../../pallets/order-book", default-features = false, features = ["private-net"] }
```
wer1st, line 26: Looks redundant because you include order-book with `private-net` feature below. It should work if you build `qa-tools` with `private-net` feature.

pallets/qa-tools/src/pallet_tools/order_book.rs
```diff
@@ -0,0 +1,368 @@
...
+fn verify_fill_side_price_params<T: Config>(
+    params: &SideFill,
+    tick_size: OrderPrice,
+) -> Result<(), DispatchError> {
+    let tick = tick_size.balance();
+    ensure!(
+        params.price_step % tick == 0
+            && params.price_step != 0
+            && params.highest_price % tick == 0
+            && params.lowest_price % tick == 0,
```
wer1st, line 189: Check:
- `highest_price` >= `lowest_price`
- `highest_price` != 0
- `lowest_price` != 0
- `orders_per_price` != 0
- `orders_per_price` <= Config::MaxLimitOrdersForPrice
- `(highest_price - lowest_price) / price_step` <= Config::MaxSidePriceCount

pallets/order-book/src/lib.rs
bragov4ik: I went with `insert` because the value is always `Some`. Why do you prefer `set`?

pallets/qa-tools/src/lib.rs
bragov4ik: I've added them before I found this out; will remove 👍

pallets/order-book/src/lib.rs
wer1st: Yes. As for me `insert` is more suitable for creation, and `set` for updating

pallets/qa-tools/Cargo.toml
wer1st: It could mean you have problems with feature flags in `qa-tools` pallet

pallets/qa-tools/src/pallet_tools/order_book.rs
```diff
@@ -0,0 +1,376 @@
...
+            && params.highest_price % tick == 0
+            && params.lowest_price % tick == 0
+            && params.highest_price != 0
+            && params.lowest_price != 0
+            && params.highest_price >= params.lowest_price
+            && params.orders_per_price != 0
+            && params.orders_per_price
+                <= <T as order_book::Config>::MaxLimitOrdersForPrice::get().into()
+            && (params.highest_price.saturating_sub(params.lowest_price) / params.price_step)
+                < <T as order_book::Config>::MaxSidePriceCount::get().into(),
```
wer1st, line 197:
```suggestion
                <= <T as order_book::Config>::MaxSidePriceCount::get().into(),
```

runtime/src/lib.rs
```diff
@@ -1456,9 +1456,9 @@ parameter_types! {
 
 #[cfg(all(feature = "private-net", feature = "ready-to-test"))] // order-book
 impl qa_tools::Config for Runtime {
+    type WeightInfo = qa_tools::weights::SubstrateWeight<Runtime>;
     type AssetInfoProvider = Assets;
     type QaToolsWhitelistCapacity = QaToolsWhitelistCapacity;
-    type WeightInfo = qa_tools::weights::SubstrateWeight<Runtime>;
```
wer1st, line 1461: I don't think we need this change.
Also, `WeightInfo` is placed at the end of configs in most cases.

pallets/order-book/Cargo.toml
```diff
@@ -67,3 +67,5 @@ runtime-benchmarks = [
 try-runtime = ["frame-support/try-runtime"]
 
 ready-to-test = ["framenode-chain-spec/ready-to-test", "framenode-runtime/ready-to-test", "common/ready-to-test"]
+
+private-net = []
```
wer1st, line 71:
```suggestion
private-net = [
    "framenode-chain-spec/private-net",
    "framenode-runtime/private-net",
]
```

pallets/qa-tools/Cargo.toml
bragov4ik: > It could mean you have problems with feature flags in `qa-tools` pallet

As far as I understand, `cargo check` includes `qa-tools` without private-net because it is included in `std` feature. This way (with `optional = true` set for the crate) it is included without the `private-net` feature, even though it remains unused. Not sure how to fix it, I can check when I have more time later;

Btw I'm always open to your suggestions to solve it as I'll be on a vacation for next 2 weeks :)

bragov4ik: I would fix it immediately, but I don't see any ways to fix it rn

runtime/src/lib.rs
bragov4ik: ok, it's intellij that complained about different order within trait and its impl. Thought it makes sense to reorder; I'll fix since we follow it everywhere.

pallets/order-book/Cargo.toml
bragov4ik: Yeah, it makes sense; I've added them previously, but this compilation error was happening with these flags in chainspec+runtime: https://jenkins.soramitsu.co.jp/blue/organizations/jenkins/sora2%2Fsubstrate/detail/PR-816/66/pipeline

need to debug reason for it and fix, agree

wer1st:
@wer1st have to check the scheduling issue

wer1st (approved):

PR #821, opened 2023-11-09
Scheduling of order-book alignment

wer1st:


pallets/order-book/src/scheduler.rs
```diff
@@ -211,3 +234,83 @@ impl<T: Config>
...
+            );
+
+            let limit_orders = Self::get_limit_orders(&order_book_id, cursor, count as usize);
+
+            if let Some(last) = limit_orders.last() {
+                new_cursors.insert(order_book_id, last.id);
+            } else {
+                // it means `limit_orders` is empty
+                finished.push(order_book);
+                continue;
```
bragov4ik, line 279: For more pedantic weight calculation, we can count iterations of the loop which are `continue`d.

To do this, the inside of the `for` loop can be segregated into smth like `service_alignment_cursor` or `service_alignment_order_book`. Then this function can be separately benchmarked on an empty case to get `align_order_book_base`, which can be subtracted at the start of the function `service_alignment_order_book`.

It shouldn't be a disaster if the weight is unaccounted; but just wanted to mention that such approach is more accurate. Especially if we consider that there are many storage accesses on each iteration.

runtime/src/lib.rs
```diff
@@ -1901,6 +1901,7 @@ impl hermes_governance_platform::Config for Runtime {
 parameter_types! {
     // small value for test environment in order to check postponing expirations
     pub ExpirationsSchedulerMaxWeight: Weight = Perbill::from_percent(15) * BlockWeights::get().max_block; // TODO: order-book clarify
+    pub AlignmentSchedulerMaxWeight: Weight = Perbill::from_percent(50) * BlockWeights::get().max_block;
```
bragov4ik, line 1904: we have substrate's scheduler `SchedulerMaxWeight` set to 80% of the block + `ExpirationsSchedulerMaxWeight` to 15%. Won't exceeding 100% (by 45%) be a problem?

wer1st: Probably, @vovac12 what di you think?

pallets/order-book/src/scheduler.rs
```diff
@@ -134,22 +137,42 @@ impl<T: Config> Pallet<T> {
...
+                .collect()
+        } else {
+            let key = <LimitOrders<T>>::hashed_key_for(order_book_id, start);
+            <LimitOrders<T>>::iter_prefix_from(order_book_id, key)
+                .take(count)
+                .map(|(_, value)| value)
+                .collect()
         }
-        postponed == 0
     }
```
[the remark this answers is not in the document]
wer1st: Ok to make `start` be `Option`. But `count` stays as it is.

pallets/order-book/src/scheduler.rs
wer1st: I prefer the current approach

bragov4ik (approved):
LGTM
just need to clarify the block weight limit part

runtime/src/lib.rs
wer1st: Decided to set `SchedulerMaxWeight` to 50% and `AlignmentSchedulerMaxWeight` to 35%.
In this case `SchedulerMaxWeight` + `ExpirationsSchedulerMaxWeight` + `AlignmentSchedulerMaxWeight` = 100%

vovac12 (approved):

PR #831, opened 2023-11-17
Enable / disable liquidity source in DexApi

wer1st:


Alexey-N-Chernyshov (approved):

runtime/src/lib.rs
```diff
@@ -256,10 +256,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("sora-substrate"),
     impl_name: create_runtime_str!("sora-substrate"),
     authoring_version: 1,
-    spec_version: 64,
+    spec_version: 69,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
-    transaction_version: 64,
+    transaction_version: 69,
```
bragov4ik, line 262: I'm not fully understanding these version numbers; should they increase by 5 like that?

bragov4ik (approved):

runtime/src/lib.rs
wer1st: master already had version 68. Next version is 69

PR #837, opened 2023-11-29
OrderBook: dynamic exchange weight calculation

bragov4ik:
- Weight calculation
  - Excluding filtered sources from calculation
  - Excluding unavailable sources from calculation
- Add variable to benchmark for better dynamic weight adjustment 
- Update execution benchmark to consider an even worse case.


common/src/traits.rs
```diff
@@ -329,15 +329,15 @@ impl<DEXId, AccountId, AssetId> LiquiditySource<DEXId, AccountId, AssetId, Fixed
...
     }
 
     fn exchange_weight() -> Weight {
-        Weight::zero()
+        Weight::from_all(2)
     }
 
     fn check_rewards_weight() -> Weight {
-        Weight::zero()
+        Weight::from_all(3)
```
wer1st, line 340: Why do you need non-zero values?

pallets/liquidity-proxy/src/lib.rs
```diff
@@ -1230,9 +1239,30 @@ impl<T: Config> Pallet<T> {
...
+                                input_asset_id,
+                                output_asset_id,
+                                &filter,
+                            )
+                            .unwrap_or(Vec::new()); // no sources -> no exchanges -> no weight
+                            let single_exchange_weight =
+                                T::LiquidityRegistry::exchange_weight_filtered(
+                                    exchange_sources.iter().map(|s| {
+                                        let b = &s.liquidity_source_index;
+                                        b
```
wer1st, line 1259:
```suggestion
                                        &s.liquidity_source_index
```

pallets/order-book/src/weights.rs
```diff
@@ -123,53 +123,43 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
...
+		//  Estimated: `21023`
+		// Minimum execution time: 43_000_000 picoseconds.
+		Weight::from_parts(49_000_000, 21023)
 			.saturating_add(T::DbWeight::get().reads(6_u64))
 			.saturating_add(T::DbWeight::get().writes(4_u64))
 	}
 	/// Storage: OrderBook OrderBooks (r:1 w:1)
-	/// Proof: OrderBook OrderBooks (max_values: None, max_size: Some(237), added: 2712, mode: MaxEncodedLen)
+	/// Proof: OrderBook OrderBooks (max_values: None, max_size: Some(238), added: 2713, mode: MaxEncodedLen)
 	/// Storage: Assets AssetInfos (r:1 w:0)
```
wer1st, line 135: Don't change previous values, just update only necessery functions.

common/src/traits.rs
bragov4ik: So that tests actually test something. I.e. if weight calculation does not add weight of some liquidity source, it is not spotted with zero weight because adding zero is no-op.

wer1st: All tests are passed with zero values.
Don't change the trait only for tests

pallets/order-book/src/lib.rs
```diff
@@ -119,10 +119,13 @@ pub mod pallet {
         const MILLISECS_PER_BLOCK: MomentOf<Self>;
         const MAX_PRICE_SHIFT: Perbill;
         /// The soft ratio between min & max order amounts.
-        /// In particular, it defines the optimal number of limit orders that could be executed by one big market order in one block.
+        /// In particular, it defines the optimal number of limit orders that could be executed by
+        /// one big market order in one block.
```
wer1st, line 123:
```suggestion
        /// In particular, it defines the optimal number of limit orders that could be executed by one big market order in one block.
```

pallets/order-book/src/lib.rs
```diff
@@ -119,10 +119,13 @@ pub mod pallet {
...
         const SOFT_MIN_MAX_RATIO: usize;
-        /// The soft ratio between min & max order amounts.
-        /// In particular, it defines the max number of limit orders that could be executed by one big market order in one block.
+        /// The hard ratio between min & max order amounts.
+        /// In particular, it defines the max number of limit orders that could be executed by
+        /// one big market order in one block.
+        /// During update of parameters, the limits must satisfy this ratio.
```
wer1st, line 128:
```suggestion
        /// The hard ratio between min & max order amounts.
        /// In particular, it defines the max number of limit orders that could be executed by one big market order in one block.
        /// During update of parameters, the limits must satisfy this ratio.
```

wer1st (approved):

PR #864, opened 2024-01-10
Liquidity source init tools

bragov4ik:
* small `BalanceUnit` improvements (trait impls + new fns + docs)
* reorganize `qa-tools` corresponding to the changes
* add initialization tools for xyk, xst
* other small fixes (mainly readability improvements)

Issue #749, opened 2023-09-29
QA Tools: initialize liquidity source


wer1st:
### As a

dev / qa

### I want to

Initialize some liquidity source for a trading pair

### So that

I can setup a necessary preconditions much faster and don't waste my time to remember all requiremenst for the liquidity source.

### Preconditions

The liquidity source for given trading pair is not initialized.

### Flow

1. Dev / QA selects the `qa-tools` pallet in polkadot.js
2. Select associated extrinsic to initialize the liquidity source by one call
3. Enter trading pair and other specific data (pool reserves, order book spread prices, etc.)
4. Call the extrinsic

### Postconditions

The liquidity source for current trading pair is initialized and available for exchange.

### Tech requirements

The main functions for liquidity source initializing must be separate from the extrinsics. They need to be available for call from the extrinsics and from other pallets. It might look like global test utils that provide shortcuts to initialize liquidity source. This will be used by qa-tools extrinsics and unit tests form other pallets.

bragov4ik:
- [x] Order book
  - [x] Tests
- [x] Xyk pool
  - [x] Tests
- [ ] XST
  - [x] Initialize base assets (synthetic base - XST & reference - DAI)
    - [ ] Tests
  - [ ] Initialize synthetics
    - [ ] Tests

PR #869, opened 2024-01-16
More flexible and precise `approx_eq`

bragov4ik:
- Add tolerances relative to magnitude of the compared numbers (i.e. when error accumulates in a larger computations)
- Add corresponding macros (to compare by constant (absolute) tolerance, by relative, or by both)
- Separate comparison logic into functions (to call if necessary outside the tests)

Used in tests for #749 (PR #864)

Inspired by https://stackoverflow.com/questions/4915462/how-should-i-do-floating-point-comparison

### DIFF
--- a/common/src/test_utils.rs
+++ b/common/src/test_utils.rs
@@ -348,29 +348,33 @@ mod test {
         ),
     ];
 
-    #[test]
-    fn should_approx_eq_match_abs_tolerance() {
-        for ApproxEqTestCase {
+    fn check_should_approx_eq(case: &ApproxEqTestCase) {
+        let ApproxEqTestCase {
             left,
             right,
             absolute_tolerance,
             relative_percentage,
-        } in APPROX_EQ_ABS_MATCH_CASES
-        {
-            let left = Fixed::from_bits(*left);
-            let right = Fixed::from_bits(*right);
-            let absolute_tolerance = Fixed::from_bits(*absolute_tolerance);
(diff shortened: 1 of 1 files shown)


Alexey-N-Chernyshov (commented):
My concern is usage of float-point approximation for fixed-point arithmetics, discussed it with @bragov4ik. Please, reassign me to the PR review in case of changes.

common/src/test_utils.rs
```diff
@@ -56,3 +62,808 @@ macro_rules! assert_noop_msg {
...
+/// behavior is expected.
+pub fn are_approx_eq_abs(
+    left: Fixed,
+    right: Fixed,
+    tolerance: Fixed,
+) -> Result<bool, ApproxEqError> {
+    if tolerance >= Fixed::ZERO {
+        Ok(are_approx_eq_abs_unchecked(left, right, tolerance))
+    } else {
+        Err(ApproxEqError::NegativeAbsoluteTolerance(tolerance))
```
Alexey-N-Chernyshov, line 93: This line is not covered. It would be nice to add a trivial test [like here](https://github.com/sora-xor/sora2-network/pull/869/files#diff-702d59a8c1fa0fa009bc2168d0ad50226dd4a6ce0e37754a86baee08185322ccR857-R869).

Alexey-N-Chernyshov (approved):
👍

wer1st (approved):

PR #879, opened 2024-01-22
Minor fixes

### LINKED ISSUE #871: [Tech]: LiquidityProxy workaround
## Problem statement
Now LiquidityProxy with insufficient liquidity returns `Math Error FixedWrapperCalculationFailed`. It should return `NotEnoughLiquidity` and this error should be handled, see #883.
Now we have workaround with an additional `LiquidityProxy::quote` call and correction of `LiquidityProxy::exchange` parameters.

## Action items
- [ ] When #833 is fixed, there is no need in workaround and additional quote should be removed.

Alexey-N-Chernyshov:
Improves error reporting.


pallets/pool-xyk/src/math.rs
```diff
@@ -192,24 +192,44 @@ impl<T: Config> Pallet<T> {
                 fxw_y_out.clone()
             };
             let nominator = fxw_x * y_out_with_fee.clone();
+            ensure!(fxw_y > y_out_with_fee, Error::<T>::NotEnoughAmountForFee);
```
bragov4ik, line 195: шt will also error in case some of the `FixedWrapper`s already contains errors (not necessarily when there's not enough amount). Seems like it's not expected here

Alexey-N-Chernyshov: I added unwrapping in the check.

pallets/pool-xyk/src/math.rs
```diff
@@ -200,13 +204,25 @@ impl<T: Config> Pallet<T> {
...
                 x_in_without_fee.clone()
             };
+            ensure!(
+                x_in.clone().get().is_ok(),
+                Error::<T>::NotEnoughLiquidityForSwap
+            );
+            ensure!(
+                x_in >= x_in_without_fee,
+                Error::<T>::NotEnoughLiquidityForSwap
+            );
```
bragov4ik, line 225: seems to be possible only if `fee_fraction` is negative (or some overflow happened, which should be caught by `ensure` right above). does this check makes sense?

bragov4ik (approved):

wer1st (approved):

N1ghtStorm (approved):

PR #889, opened 2024-02-05
Worktrees support in `build.rs` script

bragov4ik:
Building the project from non-main git worktrees resulted in error "Failed to create '.git/hooks' dir". Now instead of `.git` in current directory the `.git` from the main worktree is used, making the build possible.


runtime/build.rs
```diff
@@ -52,7 +52,16 @@ fn main() {
         "cargo:rerun-if-changed={}",
         pre_commit_hook_path.to_string_lossy()
     );
-    let enabled_hooks_dir = root_path.join(".git/hooks");
+    // Flexible path to support linked worktrees
```
Alexey-N-Chernyshov, line 55: Please, add more detailed description, it is hard to catch an issue. I would add that associated worktrees share git hooks with the main worktree and thus hooks should be installed into the main one. Probably it worth to add a link to git docs.

Alexey-N-Chernyshov (commented):
Interesting case, appreciate your effort. The bug in Jenkins clippy looks like [this one](https://github.com/rust-lang/rust/issues/50516). Probably Jenskins clippy uses very old rust.

Alexey-N-Chernyshov (commented):
Also code may be simplified a bit:
```
    let pre_commit_hook_path = PathBuf::new()
        .join(env::var("CARGO_MANIFEST_DIR").unwrap())
        .parent()
        .unwrap()
        .join(".hooks/pre-commit");
```

Alexey-N-Chernyshov:
I checked [Jenkins versions](https://jenkins.soramitsu.co.jp/blue/organizations/jenkins/sora2%2Fsubstrate/detail/PR-893/1/pipeline/):
- rust: 1.70-nightly, which seems to be `nightly-2023-03-21`
- OS: debian bullseye 11

bragov4ik:
closing since not sure how to fix & it's almost irrelevant (nobody from the team uses git worktrees in development AFAIK)

Issue #892, opened 2024-02-07
[Bug]: Fix incorrect DOT asset id in prod


bragov4ik:
### Steps to reproduce

1. Trigger the rewards for a DOT pool
2. Get the rewards

### Actual behavior

The rewards are not doubled because predefined `DOT` != actual `DOT`

### Expected behavior

The rewards are doubled

### Environment

{"label"=>"Development"}

bragov4ik:
I've investigated status of the assets in different environments (in comparison to prod):

|Asset name|predefined|dev|test|stage|
|-|-|-|-|-|
|XOR|good|good|good|good|
|DOT|incorrect id|missing|incorrect id|incorrect id|
|KSM|incorrect id|missing|incorrect id|incorrect id|
|USDT|incorrect id|missing|good|good|
|VAL|good|good|good|good|
|PSWAP|good|good|good|good|
|DAI|good|good|good|good|
|ETH|good|good|good|good|
|XSTUSD|good|good|good|good|
|XST|good|good|good|good|
|TBCD|good|good|good|good|
|KUSD|unexpectedly present|unexpectedly present|good|good|
|CERES|good|good|good|good|
|DEO|good|missing|good|good|
|HMX|good|good|missing|missing|

PR #894, opened 2024-02-07
Fix not matching asset ids

bragov4ik:
Solves #892


common/src/primitives.rs
```diff
@@ -179,6 +176,24 @@ pub const DEMETER_ASSET_ID: AssetId32<PredefinedAssetId> = AssetId32::from_bytes
 pub const HERMES_ASSET_ID: AssetId32<PredefinedAssetId> = AssetId32::from_bytes(hex!(
     "002d4e9e03f192cc33b128319a049f353db98fbf4d98f717fd0b7f66a0462142"
 ));
+#[cfg(not(feature = "private-net"))]
+pub const DOT: AssetId32<PredefinedAssetId> = AssetId32::from_bytes(hex!(
+    "0003b1dbee890acfb1b3bc12d1bb3b4295f52755423f84d1751b2545cebf000b"
+));
+#[cfg(not(feature = "private-net"))]
+pub const KSM: AssetId32<PredefinedAssetId> = AssetId32::from_bytes(hex!(
```
vovac12, line 184: Does it work for tests without `private-net` feature. We use KSM and DOT for tests in some pallets

bragov4ik: Thanks for pointing out! Just ran the test without `private-net`, and they pass successfully on my machine.
I suspect the tests use this const instead of hardcoded value, thus making them id-agnostic; everything seems fine :)

Alexey-N-Chernyshov (approved):

wer1st (approved):

PR #903, opened 2024-02-13
DOPS-2746 Add sonar, dojo

C4tWithShell:
1. [Set up sonar](https://sonar.soramitsu.co.jp/dashboard?id=sora%3Asora2-network&branch=feature%2FDOPS-2746%2Fsonar_dojo)
2. Set up dojo
3. Fixed problem with coverage, change to lcov format

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,6 +11,9 @@ def pipeline = new org.rust.AppPipeline(steps: this,
       cargoClippyCmds: ['housekeeping/clippy.sh'],
       buildTestCmds: ['housekeeping/build.sh'],
       buildArtifacts: 'framenode_runtime.compact.compressed.wasm, subwasm_report.json, pallet_list.txt',
-      pushToPublicRegistry: true
+      pushToPublicRegistry: true,
+      sonarProjectKey: 'sora:sora2-network',
+      sonarProjectName: 'sora2-network',
+      dojoProductType: 'sora'
 )
 pipeline.runPipeline()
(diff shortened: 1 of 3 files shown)


wpi86 (approved):

alexkg (approved):

Alexey-N-Chernyshov (commented):
1. ~~I don't have an access to Sonar~~, please ensure that developers can access it.
2. I was assured that DefectDojo has been already enabled: https://defectdojo.tachi.soramitsu.co.jp/product/90. At least there is the report. And I don't understand the process of Defect Dojo. We have a bunch of findings and it seems that they are ignored. I don't know who is responsible for this tool. And please, check that developers are granted an access.

Update: I got access to SonarQube.

Alexey-N-Chernyshov (changes requested):
image
1. I see the flag is set in SonarQube but I don't see any summary report on gh.

2. ```WARN: sonar.plugins.downloadOnlyRequired is false, so ALL available plugins will be downloaded``` Downloading, installing and running all the plugins for SonarQube takes time, we can enable only that we need.

3. `target` dir is analyzed, but should be ignored. Probably it is fixed in https://github.com/soramitsu/jenkins-library/pull/568 for SonarQube. The same issue for DefectDojo.

C4tWithShell:
> image
> 
> 1. I see the flag is set in SonarQube but I don't see any summary report on gh.
> 2. `WARN: sonar.plugins.downloadOnlyRequired is false, so ALL available plugins will be downloaded` Downloading, installing and running all the plugins for SonarQube takes time, we can enable only that we need.
> 3. `target` dir is analyzed, but should be ignored. Probably it is fixed in [Update sonar.groovy - do not scan target soramitsu/jenkins-library#568](https://github.com/soramitsu/jenkins-library/pull/568) for SonarQube. The same issue for DefectDojo.

1. Yes, there was an error within validation process. I fixed it
2. It's a new feature, we enabled it
3. Excluded

Cre-eD (approved):

PR #921, opened 2024-03-01
Add description field to batch swap

bragov4ik:


pallets/liquidity-proxy/src/lib.rs
```diff
@@ -2507,6 +2508,7 @@ pub mod pallet {
             max_input_amount: Balance,
             selected_source_types: Vec<LiquiditySourceType>,
             filter_mode: FilterMode,
+            additional_data: Option<BoundedVec<u8, T::MaxAdditionalDataLengthSwapTransferBatch>>,
```
wer1st, line 2511: Add the description comment above

runtime/src/lib.rs
```diff
@@ -1038,7 +1038,8 @@ parameter_types! {
     pub const MaxSubAccounts: u32 = 100;
     pub const MaxAdditionalFields: u32 = 100;
     pub const MaxRegistrars: u32 = 20;
-    pub const MaxAdditionalDataLength: u32 = 128;
+    pub const MaxAdditionalDataLengthXorlessTransfer: u32 = 128;
+    pub const MaxAdditionalDataLengthSwapTransferBatch: u32 = 2000;
```
wer1st, line 1042: 2000 looks too big

Alexey-N-Chernyshov (approved):

runtime/src/lib.rs
bragov4ik: @vovac12 can you clarify? the task mentioned 2000 bytes

Alexey-N-Chernyshov (approved):

runtime/src/lib.rs
vovac12: It's the upper limit and I think it's not so big, because it's not stored in blockchain storage and only needed to provide additional data for clients and indexers

wer1st (approved):

Issue #917, opened 2024-02-21
Partial withdrawal


Alexey-N-Chernyshov:
Decide should be partial collateral withdrawal should be available or not.
```people can only redeem their own vaults and it has to be all or nothing``` from https://github.com/sora-xor/rfps/issues/60

PR #926, opened 2024-03-06
feature(kensetsu): remove partial collateral withdrawal

Alexey-N-Chernyshov:
#917 
- Removes extrinsic `withdraw_collateral`.
- Adds PriceTools initialization to benchmarking.
- Fixes a problem when `LiquidityProxy::exchange` with `SwapAmount::with_desired_output`. The swap gives `output_amount` less than is specified: `100` was set, `99.(9)` was actually returned from the swap. To get actual out_amount swapped, the account balance before and after the swapped is checked.


pallets/kensetsu/src/lib.rs
```diff
@@ -494,56 +487,14 @@ pub mod pallet {
...
 
-        /// Withdraws collateral from a Collateralized Debt Position (CDP).
-        ///
-        /// ## Parameters
-        ///
-        /// - `origin`: The origin of the transaction.
-        /// - `cdp_id`: The ID of the CDP to withdraw collateral from.
-        /// - `collateral_amount`: The amount of collateral to withdraw.
-        #[pallet::call_index(3)]
-        #[pallet::weight(<T as Config>::WeightInfo::withdraw_collateral())]
```
wer1st, line 505: delete `withdraw_collateral()` from `weights.rs` as well

PR #939, opened 2024-03-21
fix(runtime-migration): fix KEN token AssetName

Alexey-N-Chernyshov:
`.` is not allowed in AssetName.

### DIFF
--- a/runtime/src/migrations.rs
+++ b/runtime/src/migrations.rs
@@ -60,7 +60,7 @@ impl<T: assets::Config + technical::Config> OnRuntimeUpgrade for RegisterKenToke
             assets_permissions_account_id.clone(),
             KEN.into(),
             AssetSymbol(b"KEN".to_vec()),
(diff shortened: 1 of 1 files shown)


wer1st (approved):

runtime/src/migrations.rs
```diff
@@ -60,7 +60,7 @@ impl<T: assets::Config + technical::Config> OnRuntimeUpgrade for RegisterKenToke
             assets_permissions_account_id.clone(),
             KEN.into(),
             AssetSymbol(b"KEN".to_vec()),
-            AssetName(b"Kensetsu token.".to_vec()),
+            AssetName(b"Kensetsu token".to_vec()),
```
wer1st, line 63: Change it in `develop`, not in `master`

wer1st (changes requested):

vovac12 (approved):

wer1st (approved):

PR #952, opened 2024-03-27
feature(kensetsu): adds minimal collateral limit on cdp creation

Alexey-N-Chernyshov:
Adds minimal collateral limit on cdp creation.


pallets/kensetsu/src/tests.rs
```diff
@@ -95,6 +96,24 @@ fn test_create_cdp_overflow_error() {
...
+        let minimal_balance = balance!(100);
+        set_xor_as_collateral_type(
+            Balance::MAX,
+            Perbill::from_percent(50),
+            FixedU128::from_float(0.0),
+            minimal_balance,
+        );
+
+        assert_noop!(
+            KensetsuPallet::create_cdp(alice(), XOR, balance!(0), balance!(0)),
```
wer1st, line 111: Check also that `balance!(100)` (equal with minimal balance) is ok

pallets/kensetsu/src/lib.rs
```diff
@@ -422,9 +427,13 @@ pub mod pallet {
...
-            let interest_coefficient = Self::collateral_infos(collateral_asset_id)
-                .ok_or(Error::<T>::CollateralInfoNotFound)?
-                .interest_coefficient;
+            let collateral_info = Self::collateral_infos(collateral_asset_id)
+                .ok_or(Error::<T>::CollateralInfoNotFound)?;
+            let interest_coefficient = collateral_info.interest_coefficient;
+            ensure!(
+                collateral_amount >= collateral_info.risk_parameters.minimal_collateral_deposit,
+                Error::<T>::CollateralBelowMinimal
+            );
```
wer1st, line 436: What if I call `create_cdp(some_sign, XOR, 0, 0)`?
As I see, the code processed `insert_cdp` and `deposit_event` and then doesn't call `deposit_internal` or `borrow_internal`. Is it ok?

Alexey-N-Chernyshov: Yes, it is ok. The user can deposit or borrow later with dedicated extrinsics. This methods combines deposit and borrow for the better UX. The frontend clients don't need to send 3 txs.

pallets/kensetsu/src/tests.rs
Alexey-N-Chernyshov: Tested in `test_create_cdp_sunny_day`. I changed `minimal_collateral_deposit` in the test.

pallets/kensetsu/src/lib.rs
wer1st: And what if not? User will not deposit anything later. Just creates e.g. 1000 empty CDPs and go away. Is it still ok?

Alexey-N-Chernyshov: That's why we impose this limit `minimal_collateral_deposit` which is configurable.

wer1st (approved):

stefashkaa (approved):

PR #966, opened 2024-04-04
Prepare stage release

vovac12:


Alexey-N-Chernyshov (approved):

runtime/src/lib.rs
```diff
@@ -1975,12 +1975,14 @@ impl kensetsu::Config for Runtime {
     type WeightInfo = kensetsu::weights::SubstrateWeight<Runtime>;
 }
 
+#[cfg(feature = "wip")] // Apollo
```
wer1st, line 1978: `wip` is only for dev env.
Test env & Stage env works with `ready-to-test` flag.
Do you want to enable Apollo for dev only or for Stage as well?

vovac12: Thank you for noticing

wer1st (approved):

PR #976, opened 2024-04-10
feature(kensetsu): accrue spamming protection

### LINKED ISSUE #878: Accrue calls spamming protection
## Problem statement
The beneficiary of Kensetsu wants to call accrue on every CDP as often as possible to get income from stability fee. In order to get a fee waiver for the protocol owner, accrue call done as unsigned tx. There are lots of CDPs in the system, and spamming with accrue calls could overload the system. There are protection parameter in the system named livetime of tx that prevents of double sending during some time. We need additional level of protection in accrue tx validation.
## Solution
1) It might be a timestamp of last accrue call in CDP - it is bad since requires additional storage.
2) We can sink only accrue calls that pay only fee amount larger than some threshold or make CDP unsafe. The better solution.

Alexey-N-Chernyshov:
Accrue now requires of $1 of uncollected fees to be called. It is a protection from calling unsigned tx too often.
Resolves #878


pallets/kensetsu/src/lib.rs
```diff
@@ -1149,49 +1140,76 @@ pub mod pallet {
...
             Ok(collateral_info)
         }
 
+        /// Recalculates and updates collateral interest coefficient with the current timestamp
+        fn update_collateral_interest_coefficient(
+            collateral_asset_id: &AssetIdOf<T>,
+        ) -> Result<CollateralInfo<T::Moment>, DispatchError> {
+            let updated_collateral_info =
+                Self::calculate_collateral_interest_coefficient(collateral_asset_id)?;
+            <CollateralInfos<T>>::set(collateral_asset_id, Some(updated_collateral_info.clone()));
```
wer1st, line 1191: It is called even if  `collateral_info` was not changed.
I don't know, probably Substrate checks it and doesn't write the same value, but if not - there is the unnecessary `write`.

pallets/kensetsu/src/lib.rs
```diff
@@ -1149,49 +1140,76 @@ pub mod pallet {
...
+        fn calculate_stability_fee(cdp_id: CdpId) -> Result<Balance, DispatchError> {
+            let cdp = Self::cdp(cdp_id).ok_or(Error::<T>::CDPNotFound)?;
+            let collateral_info =
+                Self::calculate_collateral_interest_coefficient(&cdp.collateral_asset_id)?;
+            let interest_coefficient = collateral_info.interest_coefficient;
+            let interest_percent = interest_coefficient
+                .checked_sub(&cdp.interest_coefficient)
+                .ok_or(Error::<T>::ArithmeticError)?
+                .checked_div(&cdp.interest_coefficient)
+                .ok_or(Error::<T>::ArithmeticError)?;
```
wer1st, line 1205: As I see, `interest_coefficient` is `FixedU128`.
Is it possible that `cdp.interest_coefficient > collateral_info.interest_coefficient` ?
If yes, it means the `checked_sub` returns error, is this intended?
If we don't need the error, probably we can use `saturating_sub` that returns 0 for this case.

wer1st: Also `checked_div` returns error for `cdp.interest_coefficient = 0`.
Do we want error or 0 as a result?

Alexey-N-Chernyshov: `interest_coefficient` always initialized with `1` and growth over time by stability percent. `collateral_info` coefficient is recalculated and assigned to cdp.interest_coefficient. Thus, `collateral_info.interest_coefficient` >= `cdp.interest_coefficient` >= 1.
If there is something wrong, we want to get an error.

pallets/kensetsu/src/lib.rs
Alexey-N-Chernyshov: Yes, good point. This write may be removed at all and calculated in runtime.

pallets/kensetsu/src/lib.rs
wer1st: Ok

wer1st (approved):

pallets/kensetsu/src/lib.rs
```diff
@@ -71,6 +71,16 @@ const VALIDATION_ERROR_CDP_SAFE: u8 = 4;
...
 const VALIDATION_ERROR_LIQUIDATION_LIMIT: u8 = 5;
 
+#[derive(
+    Debug, Clone, Encode, Decode, PartialEq, Eq, PartialOrd, Ord, Copy, scale_info::TypeInfo,
+)]
+pub enum CdpType {
+    /// Pays stability fee in underlying collateral, cannot be liquidated.
+    V1,
+    /// Pays stability fee in stable coins, can be liquidated.
+    V2,
```
wer1st, line 81: Probably you can define more descriptive names

wer1st (approved):

pallets/kensetsu/src/lib.rs
Alexey-N-Chernyshov: Yeah, it from [the proposal](https://github.com/sora-xor/rfps/issues/60#issuecomment-1913677513).

vovac12 (approved):

Issue #750, opened 2023-09-29
[Tech]: Liquidity Proxy: ALT tests


wer1st:
### Problem statement

Sora network goes ahead have several liquidity sources (almost 4). The new ALT mechanism is developed for liquidity proxy to work with all sources, but unit tests of liquidity proxy are too old and work with limited number of liquidity sources.

### Description

It is necessary to rewrite unit tests for liquidity proxy pallet.

### Definition of Done

- [x] Write new unit tests that consider new logic and all liquidity sources.
- [x] Use `Runtime` instead of `Mock`

### Requirements

* Tests check at least as much functionality/cases as they were prior to the rework.

PR #977, opened 2024-04-10
ALT: adapt liquidity proxy

wer1st:
There was an issue that is releated with old mechanism of the `SwapAmount` limit part calculation.


pallets/liquidity-proxy/src/lib.rs
```diff
@@ -544,18 +546,13 @@ impl<T: Config> Pallet<T> {
                 .into_iter()
                 .filter(|(_src, part_amount)| part_amount.amount() > balance!(0))
                 .map(|(src, part_amount)| {
-                    let part_amount = part_amount.amount();
-                    let part_limit = (FixedWrapper::from(part_amount) / amount.amount()
-                        * amount.limit())
-                    .try_into_balance()
-                    .map_err(|_| Error::CalculationError::<T>)?;
```
wer1st, line 551: This is the root cause of the problem. `part_limit` was calculated as a proportion of original `amount`.
Now ALT works with different liquidity sources, which can have very difference prices, and as a result - different limit parts. Just "proportion" doesn't work anymore.
New approach is the calculation the limit part for each liquidity source separately and provide id in `AggregatedSwapOutcome.distribution` (see liquidity_aggregator.rs).

pallets/liquidity-proxy/src/lib.rs
```diff
@@ -34,10 +34,12 @@
 
 extern crate core;
 
-mod liquidity_aggregator;
+pub mod liquidity_aggregator;
 #[cfg(test)]
 mod mock;
 #[cfg(test)]
+mod new_tests;
```
Alexey-N-Chernyshov, line 41: `new_tests` name is not descriptive, may be move the test to `tests` or give a better name?

Alexey-N-Chernyshov (approved):

pallets/liquidity-proxy/src/lib.rs
wer1st: It will be done in #750

pallets/order-book/src/test_utils/mod.rs
```diff
@@ -28,7 +28,7 @@
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-#![cfg(any(test, feature = "runtime-benchmarks"))]
+#![cfg(any(test, feature = "test", feature = "runtime-benchmarks"))]
```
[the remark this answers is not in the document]
wer1st: It allows us to not make it public. I think it is better to not make test utils be public.

vovac12 (approved):

PR #1051, opened 2024-05-21
ALT: exceed reserves

wer1st:
Process the cases when `quote` exceeds the reserves.
The main idea is that even if reserves are exceeded, `step_quote` returns as much liquidity as it can.


Alexey-N-Chernyshov (approved):

wer1st:
I have to add the irreducible balance mechanism

vovac12 (approved):

PR #1057, opened 2024-05-27
fix(kensetsu): close cdp only owner

Alexey-N-Chernyshov:
Fix `Burn others KUSD by calling repay_debt function` issue.

### DIFF
--- a/pallets/kensetsu/Cargo.toml
+++ b/pallets/kensetsu/Cargo.toml
@@ -42,6 +42,7 @@ assets = { path = "../assets", default-features = false }
 [features]
 default = ["std"]
 std = [
+	"assets/std",
 	"codec/std",
 	"frame-support/std",
 	"frame-system/std",
--- a/pallets/kensetsu/src/tests.rs
+++ b/pallets/kensetsu/src/tests.rs
@@ -914,6 +914,25 @@ fn test_repay_debt_only_signed_origin() {
     });
 }
 
+#[test]
+fn test_repay_debt_only_cdp_owner() {
+    new_test_ext().execute_with(|| {
+        set_xor_as_collateral_type(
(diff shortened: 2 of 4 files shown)


vovac12 (approved):

pallets/kensetsu/Cargo.toml
```diff
@@ -42,6 +42,7 @@ assets = { path = "../assets", default-features = false }
 [features]
 default = ["std"]
 std = [
+	"assets/std",
```
wer1st, line 45: Is it necessary if you use `assets` in `dev-dependecies` only?

pallets/kensetsu/src/tests.rs
```diff
@@ -914,6 +914,25 @@ fn test_repay_debt_only_signed_origin() {
...
+            FixedU128::from_float(0.0),
+            balance!(0),
+        );
+        // Alice is CDP owner
+        let cdp_id = create_cdp_for_xor(alice(), balance!(0), balance!(0));
+
+        assert_noop!(
+            KensetsuPallet::close_cdp(bob(), cdp_id),
+            KensetsuError::OperationNotPermitted
+        );
```
wer1st, line 932: And what about good case?

pallets/kensetsu/Cargo.toml
Alexey-N-Chernyshov: Yes, tests require `std` feature on assets.

pallets/kensetsu/src/tests.rs
Alexey-N-Chernyshov: Good cases present further, where `caller` is `alice`.

wer1st (approved):

PR #1058, opened 2024-05-27
Tech: 1053 technical replace mock config implementations with macro

ZlayaMorda:


common/src/mock.rs
```diff
@@ -327,7 +328,7 @@ macro_rules! mock_technical_config {
             type TechAccountId = TechAccountId;
             type Trigger = ();
             type Condition = ();
-            type SwapAction = ();
+            type SwapAction = pool_xyk::PolySwapAction<DEXId, AssetId, AccountId, TechAccountId>;
```
Alexey-N-Chernyshov, line 331: I'm pretty sure that most of the pallets don't use SwapAction, and unit type can be used. Moreover, if we remove dependency on `pool_xyk`, it will improve build time. But this approach requires manual checking of every usage of `mock_technical_config`.

Alexey-N-Chernyshov: I will check some pallets.

ZlayaMorda: Well, I will check

ZlayaMorda: Looks like all the pallets need it.

vovac12 (approved):

Alexey-N-Chernyshov (approved):

PR #1064, opened 2024-05-29
Tech: 1053 assets replace mock config implementations with macro

ZlayaMorda:


pallets/pool-xyk/src/mock.rs
```diff
@@ -195,6 +195,7 @@ parameter_types! {
     pub const GetBuyBackDexId: DEXId = 0;
 }
 
+// TODO: Not mocked, because of different DEX types
```
Alexey-N-Chernyshov, line 198: As I got, the problem is that `DEXId` hardcoded to only 2 values and the tests require custom DEXId. We could introduce `pub type MockDexId = u32;` in `common/mocks` for such cases.

Alexey-N-Chernyshov: The same for `pallets/liquidity-proxy/src/mock.rs`.

ZlayaMorda: Sorry, I don't understand, now it is already used as `u32` for the cases, why do we need `type MockDexId`? If we are not going to change all `DEXId` to it, it may be little bit confusing. Actually I just realized that we may:
1) remove the parameter `GetBuyBackDexId` and set up it manually;
2) set up type as macro parameter;
3) leave `GetBuyBackDexId`, but in a such way `pub GetBuyBackDexId: DEXId = DEXId::from(common::DEXId::Polkaswap);`
To my mind last one is nice enough, isn't it?

pallets/assets/src/mock.rs
```diff
@@ -116,6 +116,7 @@ parameter_types! {
     pub const GetBuyBackDexId: DEXId = DEXId::Polkaswap;
 }
 
+// TODO: Not mocked, because tests fails
```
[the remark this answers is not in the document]
Alexey-N-Chernyshov: Well, I think you should remove TODO comment, because actually there is nothing to be done =)

Alexey-N-Chernyshov (approved):

pallets/eth-bridge/src/tests/cancel.rs
```diff
@@ -47,7 +47,8 @@ use crate::tests::{
 };
 use crate::{AssetConfig, EthAddress};
 use common::{
-    AssetInfoProvider, AssetName, AssetSymbol, DEFAULT_BALANCE_PRECISION, DOT, KSM, USDT, VAL, XOR,
+    AssetInfoProvider, AssetName, AssetSymbol, DEFAULT_BALANCE_PRECISION, DOT, KSM, PSWAP, USDT,
+    VAL, XOR,
```
wer1st, line 51: Why do you add new assets in `use`? And in other files.
I don't see they are used here.

ZlayaMorda: In the pallet it was implemented as constant, which duplicate common const, so I removed it and import. And maybe somewhere it used manually (path::const) and I changed to import. Anyway it is auto checked and unused import won't exist.

wer1st (approved):

Issue #1071, opened 2024-06-04
[Tech]: Chameleon pool quote limitations


vovac12:
### Problem statement

Quote/step_quote should return error if swap is not possible

### Definition of Done

- [ ] Check output asset amount in quote
- [ ] Return correct limits in step_quote

PR #1078, opened 2024-06-07
XYK Chameleon pool quote limits

vovac12:
#1071


pallets/pool-xyk/src/lib.rs
```diff
@@ -461,16 +469,30 @@ impl<T: Config> LiquiditySource<T::DEXId, T::AccountId, AssetIdOf<T>, Balance, D
...
+                };
+                ensure!(
+                    required_output <= max_output_available,
+                    Error::<T>::NotEnoughOutputReserves
+                );
+                (calculated, fee_amount)
             }
             QuoteAmount::WithDesiredOutput { desired_amount_out } => {
+                ensure!(
+                    desired_amount_out <= max_output_available,
```
wer1st, line 493: In case when `max_output_available` == `reserve_output` == `desired_amount_out` it will fail when divided by 0 in `calc_input_for_exact_output()`.
Probably just `desired_amount_out < max_output_available` is better.

Also take a note that `calc_input_for_exact_output()` has a line:
```rust
let fxw_y_out = fxw_y_out.clone() + Fixed::from_bits(1);
```

pallets/pool-xyk/src/tests.rs
```diff
@@ -3821,3 +3821,249 @@ fn initialize_pool_with_synthetics() {
...
+                        balance!(16.442360950037693547),
+                        OutcomeFee::from_asset(GoldenTicket.into(), balance!(0.049475509378247824))
+                    ),
+                    SwapChunk::new(
+                        balance!(33),
+                        balance!(16.436938197461963666),
+                        OutcomeFee::from_asset(GoldenTicket.into(), balance!(0.049459192168892569))
+                    ),
+                ]),
+                limits: SwapLimits::new(None, None, None)
```
wer1st, line 3878: Easier
```suggestion
                limits: Default::default()
```

pallets/pool-xyk/src/lib.rs
```diff
@@ -461,16 +470,30 @@ impl<T: Config> LiquiditySource<T::DEXId, T::AccountId, AssetIdOf<T>, Balance, D
...
+                };
+                ensure!(
+                    required_output <= max_output_available,
+                    Error::<T>::NotEnoughOutputReserves
+                );
+                (calculated, fee_amount)
             }
             QuoteAmount::WithDesiredOutput { desired_amount_out } => {
+                ensure!(
+                    desired_amount_out < max_output_available,
```
wer1st, line 494: It doesn't solve the problem completely.

`calc_input_for_exact_output` has a line:
```rust
let fxw_y_out = fxw_y_out.clone() + Fixed::from_bits(1);
```

It means if `desired_amount_out` == `max_output_available - 1`, then
`desired_amount_out < max_output_available` is true, but inside of `calc_input_for_exact_output` it fails when divided by 0.

vovac12: I think it's part of initial XYK pool logic and will just return an error. This check is more about checking that in Chameleon pool we have enough required asset

wer1st: Ok

wer1st (approved):

Alexey-N-Chernyshov (approved):

PR #1082, opened 2024-06-10
[Tech]: 1068 dynamic fee update

ZlayaMorda:


pallets/xor-fee/src/lib.rs
```diff
@@ -695,7 +732,33 @@ pub mod pallet {
...
+                ) {
+                    Ok(new_multiplier) => {
+                        <Multiplier<T>>::put(new_multiplier); // 1 write
+                        Self::deposit_event(Event::WeightToFeeMultiplierUpdated(new_multiplier)); // 1 write
+
+                        let next_update_block = current_block + T::BlocksToUpdate::get(); // 1 read
+                        <NextUpdateBlock<T>>::put(next_update_block); // 1 write
+                        weight += T::DbWeight::get().reads(2) + T::DbWeight::get().writes(3);
+                    }
+                    Err(e) => {
```
ZlayaMorda, line 753: I'm not sure if we have to count the multiplier every block in case of an error?
Or we may set new `NextUpdateBlock`, maybe smaller than in Ok case.

vovac12:
1. I think extrinsic should be something like `set_fee_update_period(period: Option<Blocknumber>)`. And if period is `None` then we turn off dynamic fee updates
2. Maybe it's better to use `on_initialize` hook with `now % period == 0` check, so we don't need to have one more custom scheduler and storage

ZlayaMorda:
> 1. I think extrinsic should be something like `set_fee_update_period(period: Option<Blocknumber>)`. And if period is `None` then we turn off dynamic fee updates
> 2. Maybe it's better to use `on_initialize` hook with `now % period == 0` check, so we don't need to have one more custom scheduler and storage

1. Yeah, it is nice
2. It is the way too, but it would reduce the flexibility a little bit. We have one storage for period anyway. I think it is not really necessary to update multiplier right at the block, I mean update period may be a bit more. So we may use idle time and leftover weight.

runtime/src/lib.rs
```diff
@@ -1299,9 +1299,15 @@ where
...
 parameter_types! {
     pub const DEXIdValue: DEXId = 0;
+    pub const BlocksToUpdate: BlockNumber = (MILLISECS_PER_MULTIPLIER_UPDATE / MILLISECS_PER_BLOCK) as BlockNumber;
 }
 
 impl xor_fee::Config for Runtime {
+    type BlocksToUpdate = BlocksToUpdate;
+    #[cfg(not(feature = "ready-to-test"))] // Dynamic fee
+    type DynamicMultiplier = ();
+    #[cfg(feature = "ready-to-test")] // Dynamic fee
```
wer1st, line 1309: Do we want to deploy it on test & stage envs immediately? Why not `wip`?

pallets/xor-fee/src/lib.rs
```diff
@@ -428,6 +429,39 @@ pub trait WithdrawFee<T: Config> {
...
+        Err(DispatchError::CannotLookup)
+    }
+
+    fn calculate_multiplier(
+        _input_asset: &AssetId,
+        _ref_asset: &AssetId,
+    ) -> Result<FixedU128, DispatchError> {
+        Err(DispatchError::CannotLookup)
+    }
+}
```
wer1st, line 463: Probably we can move it in `common/traits.rs`

pallets/xor-fee/src/lib.rs
```diff
@@ -695,7 +733,40 @@ pub mod pallet {
...
+                        &common::XOR.into(),
+                        &common::DAI.into(),
+                    ) {
+                        Ok(new_multiplier) => {
+                            <Multiplier<T>>::put(new_multiplier); // 1 write
+                            Self::deposit_event(Event::WeightToFeeMultiplierUpdated(
+                                new_multiplier,
+                            ));
+
+                            let next_block_number = current_block + T::BlocksToUpdate::get(); // 1 read
```
wer1st, line 753: It's better to use `checked_add` or `saturating_add` wherever we can to avoid even the ability to panic.

pallets/xor-fee/src/lib.rs
```diff
@@ -714,6 +785,22 @@ pub mod pallet {
...
+
+        #[allow(unused_variables)]
+        #[pallet::call_index(1)]
+        #[pallet::weight(<T as Config>::WeightInfo::set_fee_update_period())]
+        pub fn set_fee_update_period(
+            origin: OriginFor<T>,
+            next_block_number: Option<<T as frame_system::Config>::BlockNumber>,
+        ) -> DispatchResultWithPostInfo {
+            ensure_root(origin)?;
+            #[cfg(feature = "ready-to-test")] // Dynamic fee
```
wer1st, line 797:
```suggestion
        #[cfg(feature = "wip")] // Dynamic fee
        #[pallet::call_index(1)]
        #[pallet::weight(<T as Config>::WeightInfo::set_fee_update_period())]
        pub fn set_fee_update_period(
            origin: OriginFor<T>,
            next_block_number: Option<<T as frame_system::Config>::BlockNumber>,
        ) -> DispatchResultWithPostInfo {
            ensure_root(origin)?;
```

pallets/xor-fee/src/lib.rs
```diff
@@ -714,6 +785,22 @@ pub mod pallet {
...
+            {
+                <NextUpdateBlock<T>>::put(next_block_number);
```
wer1st, line 799: Probably you misunderstand the task. `BlocksToUpdate` should not be a config param, it shoud be a storage.
And this extrinsic should not just update the next block where hook is triggered, it should update this period.

pallets/xor-fee/src/lib.rs
```diff
@@ -695,7 +733,40 @@ pub mod pallet {
     pub struct Pallet<T>(PhantomData<T>);
 
     #[pallet::hooks]
-    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        #[cfg(feature = "ready-to-test")] // Dynamic fee
+        fn on_idle(current_block: BlockNumberFor<T>, _remaining_weight: Weight) -> Weight {
+            let mut weight: Weight = Weight::default();
+            if let Some(next_update_block) = Self::next_update_block() {
+                if current_block >= next_update_block {
```
wer1st, line 741: You have to check that `BlocksToUpdate != 0`, because if it is 0 - it means the tech committee turned off the dynamic update.
But now, if the value is 0, the hook will trigger each block because `next_block_number == current_block`

pallets/xor-fee/src/lib.rs
```diff
@@ -714,6 +785,22 @@ pub mod pallet {
...
         }
+
+        #[allow(unused_variables)]
+        #[pallet::call_index(1)]
+        #[pallet::weight(<T as Config>::WeightInfo::set_fee_update_period())]
+        pub fn set_fee_update_period(
+            origin: OriginFor<T>,
+            next_block_number: Option<<T as frame_system::Config>::BlockNumber>,
+        ) -> DispatchResultWithPostInfo {
+            ensure_root(origin)?;
```
wer1st, line 796: It checks the democracy or sudo, but you need to check the tech committee credentials.

pallets/xor-fee/src/lib.rs
```diff
@@ -714,6 +785,22 @@ pub mod pallet {
...
+            #[cfg(feature = "ready-to-test")] // Dynamic fee
+            {
+                <NextUpdateBlock<T>>::put(next_block_number);
+                Self::deposit_event(Event::NextUpdateBlockUpdated(next_block_number));
```
wer1st, line 800: Also we need another event for the period update.

pallets/xor-fee/src/lib.rs
wer1st: Or, as @vovac12 mentions above, you have to check that period is not `None` if it will be an `Option`.
However, you still need to check zero case. And add min possible period check probably.

pallets/xor-fee/src/lib.rs
ZlayaMorda: We may, but it is used only there and I don't know where it may be used else

pallets/xor-fee/src/lib.rs
ZlayaMorda: If 0 is not an off number, we also must check it not be set via extrinsic.

ZlayaMorda: > And add min possible period check probably.

Maybe max too?

wer1st: Maybe. It's necessary to clarify requirements.

runtime/src/xor_fee_impls.rs
```diff
@@ -29,13 +29,22 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use crate::*;
+#[cfg(feature = "ready-to-test")] // Dynamic fee
+use _feps::sp_arithmetic::FixedU128;
```
Alexey-N-Chernyshov, line 33:
```suggestion
use sp_runtime::FixedU128;
```

common/src/weights.rs
```diff
@@ -50,6 +50,7 @@ pub mod constants {
     pub const EXTRINSIC_FIXED_WEIGHT: Weight = Weight::from_parts(100_000_000, 0);
     pub const SMALL_FEE: Balance = balance!(0.0007);
     pub const BIG_FEE: Balance = balance!(0.007);
+    pub const SMALL_REFERENCE_AMOUNT: Balance = balance!(0.2);
```
Alexey-N-Chernyshov, line 53: Please, add comment that small fee value should be `0.2` in reference asset id (DAI).

pallets/xor-fee/src/lib.rs
```diff
@@ -428,6 +429,39 @@ pub trait WithdrawFee<T: Config> {
     ) -> Result<(T::AccountId, Option<NegativeImbalanceOf<T>>), DispatchError>;
 }
 
+/// Trait for dynamic fee update via multiplier
+pub trait CalculateMultiplier<AssetId, Error> {
+    /// Parameters:
+    /// `input_asset` is asset id which price should be fetched;
+    /// `ref_asset` is asset id in which price will be fetched
+    fn fetch_price_to_reference_asset(
```
Alexey-N-Chernyshov, line 437: I don't see that the method is used somewhere except `calculate_multiplier`. You can remove it and keep  only `calculate_multiplier` in the trait.

pallets/xor-fee/src/lib.rs
```diff
@@ -714,6 +785,22 @@ pub mod pallet {
             Self::deposit_event(Event::WeightToFeeMultiplierUpdated(new_multiplier));
             Ok(().into())
         }
+
+        #[allow(unused_variables)]
```
Alexey-N-Chernyshov, line 782:
```suggestion
```

pallets/xor-fee/src/migrations.rs
```diff
@@ -0,0 +1,81 @@
...
+    use crate::*;
+
+    pub struct Migrate<T>(PhantomData<T>);
+
+    impl<T> OnRuntimeUpgrade for Migrate<T>
+    where
+        T: Config,
+    {
+        fn on_runtime_upgrade() -> Weight {
+            if StorageVersion::get::<Pallet<T>>() < StorageVersion::new(2) {
```
Alexey-N-Chernyshov, line 49: Please, update storage version in `pallets/xor-fee/src/lib.rs:727`.

pallets/xor-fee/src/lib.rs
```diff
@@ -726,7 +813,24 @@ pub mod pallet {
...
+        MultiplierCalculationFailed,
+    }
+
+    #[cfg(feature = "ready-to-test")] // Dynamic fee
+    /// Next block number to update multiplier
+    /// If it is necessary to stop updating the multiplier,
+    /// set BlockNumber::MAX or another large value
+    #[pallet::storage]
+    #[pallet::getter(fn next_update_block)]
+    pub type NextUpdateBlock<T> =
```
Alexey-N-Chernyshov: It is better to use `LastUpdateBlock` and check `current_block >= last_update_block + update_period`. In this case if you change parameter `update_period`, you don't need to change the `NextUpdateBlock`.

common/src/weights.rs
ZlayaMorda: @vovac12 Maybe the value must be storage and updateable by extrinsic too?

pallets/xor-fee/src/lib.rs
ZlayaMorda: Decided to use only period

common/src/weights.rs
vovac12: Good idea. Let's do it like that

pallets/xor-fee/src/lib.rs
vovac12: I think we can just turn off on 0, so min check is not necessary. And I don't think that we need to check for the max value

pallets/xor-fee/src/lib.rs
```diff
@@ -714,6 +758,36 @@ pub mod pallet {
...
+            {
+                <UpdatePeriod<T>>::put(new_period);
+                Self::deposit_event(Event::PeriodUpdated(new_period));
+            }
+            Ok(().into())
+        }
+
+        #[pallet::call_index(2)]
+        #[pallet::weight(<T as Config>::WeightInfo::set_small_reference_amount())]
+        pub fn set_small_reference_amount(
```
wer1st, line 779: Add the `///` descriptions for new extrinsics

pallets/xor-fee/src/lib.rs
```diff
@@ -695,7 +717,29 @@ pub mod pallet {
     pub struct Pallet<T>(PhantomData<T>);
 
     #[pallet::hooks]
-    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        #[cfg(feature = "wip")] // Dynamic fee
+        fn on_initialize(current_block: BlockNumberFor<T>) -> Weight {
+            let update_period = Self::update_period(); // 1 read
+            let mut weight: Weight = T::DbWeight::get().reads(1);
+            if update_period != 0_u32.into() && current_block % update_period == 0_u32.into() {
```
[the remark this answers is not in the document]
ZlayaMorda: I didn't see it, thank you.

pallets/xor-fee/src/lib.rs
```diff
@@ -714,6 +758,36 @@ pub mod pallet {
...
+                Self::deposit_event(Event::PeriodUpdated(new_period));
+            }
+            Ok(().into())
+        }
+
+        #[pallet::call_index(2)]
+        #[pallet::weight(<T as Config>::WeightInfo::set_small_reference_amount())]
+        pub fn set_small_reference_amount(
+            origin: OriginFor<T>,
+            new_reference_amount: Balance,
```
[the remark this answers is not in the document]
ZlayaMorda: Yeah, make sense.

runtime/src/lib.rs
```diff
@@ -1302,6 +1302,18 @@ parameter_types! {
...
 
 impl xor_fee::Config for Runtime {
+    type PermittedSetPeriod = EitherOfDiverse<
+        pallet_collective::EnsureProportionMoreThan<AccountId, TechnicalCollective, 1, 2>,
+        EnsureRoot<AccountId>,
+    >;
+    type PermittedSetSmallReferenceAmount = EitherOfDiverse<
+        pallet_collective::EnsureProportionMoreThan<AccountId, TechnicalCollective, 2, 3>,
+        EnsureRoot<AccountId>,
+    >;
```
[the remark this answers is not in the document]
ZlayaMorda: I think that changing `SmallReferenceAmount` may cause the network to fail more likely if the wrong value is set. While changing the period will break the update in the worst case scenario.
So for the first one setting I set higher proportion.

pallets/xor-fee/src/tests.rs
```diff
@@ -497,3 +505,118 @@ fn it_works_referrer_refund() {
...
+        System::set_block_number(0);
+
+        UpdatePeriod::<Runtime>::put(10);
+        XorFee::on_initialize(9);
+        assert_eq!(Multiplier::<Runtime>::get(), FixedU128::from(600000));
+
+        XorFee::on_initialize(20);
+        assert_eq!(
+            Multiplier::<Runtime>::get().into_inner(),
+            3571428571428571428571428
```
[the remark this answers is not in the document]
ZlayaMorda: Yeah, cause default value 0.2 was divided by 0.0007, and only price changing do not correct situation. Anyway changed mock values for more accurate values.

pallets/xor-fee/src/tests.rs
```diff
@@ -497,3 +505,118 @@ fn it_works_referrer_refund() {
...
+            RuntimeOrigin::root(),
+            BlockNumber::MAX
+        ));
+        assert_eq!(UpdatePeriod::<Runtime>::get(), BlockNumber::MAX);
+    });
+}
+
+#[cfg(feature = "wip")] // Dynamic fee
+#[test]
+fn non_root_set_period_fails() {
```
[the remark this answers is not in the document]
ZlayaMorda: Not sure that I should test it at all. And testing collective requires `pallet_collective`, which would be extra import here. So removed them.

runtime/src/lib.rs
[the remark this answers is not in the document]
wer1st: @vovac12 please clarify the requirements for tech committee vote proportions.

pallets/xor-fee/src/tests.rs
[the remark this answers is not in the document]
wer1st: `pallet-collective` could be added in `dev-dependencies`, but `xor-fee` contains bencharks inside.
Benchmarks use the same `mock` and it requres add `pallet-collective` as a regular dependency that is not good.
It's why better to move benchmarks to the separate crate.

Ok, you can skip this test case here, but you have to control that QA will test it with tech committee voting as well, not only with root.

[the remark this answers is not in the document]
vovac12: I don't think we should test that `EnsureOrigin` works properly, that's out of scope and tested by Parity. It's enough to test that proper update origin is checked, for example by defining mock `EnsureOrigin` implementation

runtime/src/lib.rs
```diff
@@ -1302,6 +1302,14 @@ parameter_types! {
 }
 
 impl xor_fee::Config for Runtime {
+    type PermittedSetPeriod = EitherOfDiverse<
+        pallet_collective::EnsureProportionMoreThan<AccountId, TechnicalCollective, 2, 3>,
```
wer1st, line 1306:
```suggestion
        pallet_collective::EnsureProportionAtLeast<AccountId, TechnicalCollective, 3, 4>,
```

Alexey-N-Chernyshov (approved):

PR #1086, opened 2024-06-10
1013/expirable soulbound tokens support

HaidarJbeily7:
### Changes

- [x] Add SBT tokens support to Regulated assets pallet
     - Added more fields`SoulboundTokenMetdata` concerning Data Model in tech requirement.
     - Added `update_sbt_expiration` extrinsic. (This requirement is mentioned [HERE](https://soramitsu.atlassian.net/wiki/spaces/SE/pages/4415750152?focusedCommentId=4605083989))
     - Modified `issue_sbt` extrinsic with the new fields, added validation to `allowed_assets` and `expires_at`, and reverted the `ISSUE_SBT` permission checking.
     - Added check in `regulate_asset` extrinsic to not allow to regulate SBT.
- [x] Deny transfers for SBT tokens
     - Added this logic in `AssetRegulator.check_permission(..)`
- [x] Add expiration to SBT tokens
     - The `expires_at` field is optional, which achieves the optionality of expirable support.
     - Added expiration checking logic in `AssetRegulator.check_permission(..)`
     
* Thorough unit tests have been added to validate the new functionalities. These tests cover a wide range of scenarios to ensure robustness and correctness.


Alexey-N-Chernyshov (approved):

pallets/regulated-assets/src/lib.rs
```diff
@@ -154,37 +171,56 @@ pub mod pallet {
...
-                None,
+                image.clone(),
                 description.clone(),
             )?;
+
             let metadata = SoulboundTokenMetadata {
+                owner: who.clone(),
                 name,
                 description,
+                image: image.clone(),
```
vovac12, line 220: Do we really need to duplicate owner, image and description here? We already have it in AssetInfo

vovac12:
Should expiration be applied per-user?

pallets/regulated-assets/src/lib.rs
HaidarJbeily7: No, we don't need it. I thought of the case to ease retrieving data, but I was wrong.
I will remove name, description, image, and description fields to not duplicate info.

ok? @vovac12

HaidarJbeily7:
@vovac12 I modified the storage layout with your suggestions and modified the logic respectively.

pallets/regulated-assets/src/lib.rs
```diff
@@ -154,55 +154,121 @@ pub mod pallet {
...
         #[pallet::weight(<T as Config>::WeightInfo::issue_sbt())]
         pub fn issue_sbt(
             origin: OriginFor<T>,
             symbol: AssetSymbol,
             name: AssetName,
-            allowed_assets: BoundedVec<AssetIdOf<T>, T::MaxAllowedTokensPerSBT>,
             description: Option<Description>,
+            image: Option<ContentSource>,
+            external_url: Option<ContentSource>,
+            allowed_assets: BoundedVec<AssetIdOf<T>, T::MaxAllowedAssetsPerSBT>,
```
vovac12, line 172: I think we can create separate extrinsic to bound SBT asset with regulated one. So regulator can add new regulated assets and during setup regulator just can pass several calls of new extrinsic in `Utility.batch`

HaidarJbeily7: @vovac12 Done

vovac12 (approved):

Alexey-N-Chernyshov (approved):

PR #1094, opened 2024-06-13
Feature/apollo update

DanijelCBS:
- Added loan_to_value parameter in borrow extrinsic to make loan to value flexible not fixed
- Implemented add_collateral extrinsic for users to add collateral and strengthen their borrowing position
- Added tests for those cases

### DIFF
--- a/pallets/apollo-platform/src/weights.rs
+++ b/pallets/apollo-platform/src/weights.rs
@@ -45,6 +45,7 @@ pub trait WeightInfo {
 	fn liquidate() -> Weight;
 	fn remove_pool() -> Weight;
 	fn edit_pool_info() -> Weight;
+	fn add_collateral() -> Weight;
 }
 
 /// Weight functions for `apollo_platform`.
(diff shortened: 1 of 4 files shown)


pallets/apollo-platform/src/weights.rs
```diff
@@ -532,4 +548,19 @@ impl WeightInfo for () {
...
+	fn add_collateral() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `1210`
+		//  Estimated: `13530`
+		// Minimum execution time: 21_791 nanoseconds.
+		Weight::from_parts(22_489_000, 13530)
+			.saturating_add(RocksDbWeight::get().reads(4))
+			.saturating_add(RocksDbWeight::get().writes(3))
+	}
 }
```
Alexey-N-Chernyshov, line 566:
```suggestion
}

```

Alexey-N-Chernyshov (approved):

Tieumsan:
> * Added loan_to_value parameter in borrow extrinsic to make loan to value flexible not fixed

@DanijelCBS can you help me understand why you want the Loan To Value be flexible please ?

vovac12 (approved):

DanijelCBS:
> > * Added loan_to_value parameter in borrow extrinsic to make loan to value flexible not fixed
> 
> @DanijelCBS can you help me understand why you want the Loan To Value be flexible please ?

The Loan to Value (“LTV”) ratio defines the maximum amount of assets that can be borrowed with a specific collateral. It is expressed as a percentage (e.g., at LTV=75%, for every 1 ETH worth of collateral, borrowers will be able to borrow 0.75 ETH worth of the corresponding currency.

Each pool has a predefined loan-to-value (LTV) ratio, which currently restricts users to borrowing based solely on this fixed ratio. This approach presents drawbacks since the predefined LTV represents the maximum allowable for the pool, potentially resulting in a lower health factor and increased liquidation risk. Instead, the defined LTV should serve as a ceiling, allowing users flexibility to choose their preferred LTV. Opting for a lower LTV would result in reduced borrowing but also lower liquidation risk and a higher health factor.

DanijelCBS:
> > > * Added loan_to_value parameter in borrow extrinsic to make loan to value flexible not fixed
> > 
> > 
> > @DanijelCBS can you help me understand why you want the Loan To Value be flexible please ?
> 
> The Loan to Value (“LTV”) ratio defines the maximum amount of assets that can be borrowed with a specific collateral. It is expressed as a percentage (e.g., at LTV=75%, for every 1 ETH worth of collateral, borrowers will be able to borrow 0.75 ETH worth of the corresponding currency.
> 
> Each pool has a predefined loan-to-value (LTV) ratio, which currently restricts users to borrowing based solely on this fixed ratio. This approach presents drawbacks since the predefined LTV represents the maximum allowable for the pool, potentially resulting in a lower health factor and increased liquidation risk. Instead, the defined LTV should serve as a ceiling, allowing users flexibility to choose their preferred LTV. Opting for a lower LTV would result in reduced borrowing but also lower liquidation risk and a higher health factor.

At the moment when user borrows it is always the riskiest borrow because LTV is fixed to the max LTV for the pool.

Tieumsan:
@DanijelCBS that's clear, thanks

PR #1105, opened 2024-06-20
1015/permissioned pools support for xyk pool and liquidityproxy

HaidarJbeily7:
## Changes

- Added `AssetRegulator` as an associated type to PoolXYK pallet
- Added `check_permission` with `AssetRegulator` on `deposit_liquidity`, `withdraw_liquidity` , and `exchange`
- Added unit tests to check if all pool operations is working fine with new permission logic
- Verified `liquidity_proxy` logic is working fine with `permissioned pools` by testing `swap` with its different variants (multistep, one step, desired_output, desired_input)


pallets/liquidity-proxy/src/mock.rs
```diff
@@ -925,6 +940,54 @@ impl ExtBuilder {
...
+            None,
+        )
+        .expect("Failed to issue SBT");
+
+        // Extract the issued SBT asset ID
+        let event = frame_system::Pallet::<Runtime>::events()
+            .pop()
+            .expect("Expected at least one event")
+            .event;
+        let sbt_asset_id = match event {
```
vovac12, line 975: You can add public function to extended assets pallet to register sbt, return asset id from this function and use it here

vovac12 (approved):

pallets/extended-assets/src/tests.rs
```diff
@@ -30,58 +30,14 @@
 
 #![cfg(feature = "wip")] // DEFI-R
 
-use crate::mock::*;
+use crate::mock::{Timestamp, *};
```
ZlayaMorda, line 33: * should import all, shouldn't?

ZlayaMorda (approved):

pallets/extended-assets/src/tests.rs
HaidarJbeily7: It should, but need to be explicitly importing Timestamp to differ the mock type from the one declared in lib.rs.

PR #1115, opened 2024-07-02
1112/add storage to store regulated-assets for SBT

HaidarJbeily7:
## Changes

- Added `regulated_assets` as `BTreeSet` to `SoulboundTokenMetadata`
- Manage this set when binding regulated asset to SBT in `bind_regulated_asset_to_sbt` extrinsic
- Added unit test for testing the logic
- Ran benchmarks and updated the weights

### DIFF
--- a/pallets/extended-assets/src/lib.rs
+++ b/pallets/extended-assets/src/lib.rs
@@ -297,14 +297,15 @@ pub mod pallet {
 
             // Bind the regulated asset to the SBT and update the SBT's metadata
             <RegulatedAssetToSoulboundAsset<T>>::set(regulated_asset_id, sbt_asset_id);
-            <SoulboundAsset<T>>::mutate(sbt_asset_id, |metadata| {
+            <SoulboundAsset<T>>::try_mutate(sbt_asset_id, |metadata| -> DispatchResult {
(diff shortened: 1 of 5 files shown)


pallets/extended-assets/src/lib.rs
```diff
@@ -274,7 +284,27 @@ pub mod pallet {
...
+            }
+
+            // Bind the regulated asset to the SBT and update the SBT's metadata
             <RegulatedAssetToSoulboundAsset<T>>::set(regulated_asset_id, sbt_asset_id);
+            <SoulboundAsset<T>>::mutate(sbt_asset_id, |metadata| {
+                if let Some(metadata) = metadata {
+                    metadata
+                        .regulated_assets
+                        .try_insert(regulated_asset_id)
+                        .ok();
```
vovac12, line 305: I think we should return error if something went wrong

vovac12 (approved):

pallets/extended-assets/src/lib.rs
HaidarJbeily7: @vovac12 Handled, please check.

vovac12 (approved):

ZlayaMorda (approved):

PR #1143, opened 2024-07-23
1135/DeFi-R. User Flow Optimization: Two Flows Support

HaidarJbeily7:
## Changes

- Added regulate_asset extrinsic with benchmarking and unit tests

### DIFF
--- a/pallets/extended-assets/src/lib.rs
+++ b/pallets/extended-assets/src/lib.rs
@@ -341,6 +341,35 @@ pub mod pallet {
 
             Ok(())
         }
+
+        /// Marks an asset as regulated, representing that the asset will only operate between KYC-verified wallets.
+        ///
+        /// ## Parameters
+        ///
+        /// - `origin`: The origin of the transaction.
+        /// - `asset_id`: The identifier of the asset.
+        #[pallet::call_index(4)]
+        #[pallet::weight(<T as Config>::WeightInfo::regulate_asset())]
+        pub fn regulate_asset(origin: OriginFor<T>, asset_id: AssetIdOf<T>) -> DispatchResult {
+            let who = ensure_signed(origin)?;
+            <T as Config>::AssetInfoProvider::ensure_asset_exists(&asset_id)?;
+            ensure!(
+                <T as Config>::AssetInfoProvider::is_asset_owner(&asset_id, &who),
+                <Error<T>>::OnlyAssetOwnerCanRegulate
+            );
+            ensure!(
+                !Self::is_asset_regulated(&asset_id),
+                <Error<T>>::AssetAlreadyRegulated
+            );
+            ensure!(
+                Self::soulbound_asset(asset_id).is_none(),
+                <Error<T>>::NotAllowedToRegulateSoulboundAsset
+            );
+
+            T::AssetManager::update_asset_type(&asset_id, &AssetType::Regulated)?;
+            Self::deposit_event(Event::AssetRegulated { asset_id });
+            Ok(())
+        }
     }
 
     #[pallet::event]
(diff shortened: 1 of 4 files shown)


pallets/extended-assets/src/lib.rs
```diff
@@ -341,6 +341,35 @@ pub mod pallet {
...
+        /// Marks an asset as regulated, representing that the asset will only operate between KYC-verified wallets.
+        ///
+        /// ## Parameters
+        ///
+        /// - `origin`: The origin of the transaction.
+        /// - `asset_id`: The identifier of the asset.
+        #[pallet::call_index(4)]
+        #[pallet::weight(<T as Config>::WeightInfo::regulate_asset())]
+        pub fn regulate_asset(origin: OriginFor<T>, asset_id: AssetIdOf<T>) -> DispatchResult {
+            let who = ensure_signed(origin)?;
```
wer1st, line 354: Do we want to allow this for `sudo` as well?

pallets/extended-assets/src/lib.rs
```diff
@@ -341,6 +341,35 @@ pub mod pallet {
...
+
+        /// Marks an asset as regulated, representing that the asset will only operate between KYC-verified wallets.
+        ///
+        /// ## Parameters
+        ///
+        /// - `origin`: The origin of the transaction.
+        /// - `asset_id`: The identifier of the asset.
+        #[pallet::call_index(4)]
+        #[pallet::weight(<T as Config>::WeightInfo::regulate_asset())]
+        pub fn regulate_asset(origin: OriginFor<T>, asset_id: AssetIdOf<T>) -> DispatchResult {
```
wer1st, line 353: As I remember, the main idea is to allow to operate with the asset only for KYC passed users.
What if the asset is registered and distributed between non-KYC users and then after that we will regulate the asset?
In this case we have non-KYC users with regulated asset.

pallets/extended-assets/src/lib.rs
HaidarJbeily7: There is no such requirement so, I don't think so. it's just in the scope of asset owner.

pallets/extended-assets/src/lib.rs
HaidarJbeily7: Yes, it's possible for asset owner to regulate the asset after distribution.
Simply enough they can force the users to verify KYC by regulating the asset in order to operate further.

vovac12 (approved):

pallets/extended-assets/src/lib.rs
wer1st: We have to address the question to requirements.

HaidarJbeily7: @AndreyKortosov 
Can you please address this in requirements?

The question:
Do we need the sudo (root) account to have access to regulate any asset in addition to AssetOwner account?


cc. @wer1st

AndreyKortosov: @HaidarJbeily7 AssetOwner account will regulate any asset. Am I right?
What's the purpose of sharing access to another account?

HaidarJbeily7: @AndreyKortosov 
AssetOwner account will be able to regulate **the assets that the account owns**.
There is a sudo account which has superior accesses. So, we need to address in requirements if we need to share access with this sudo account to this extrinsic, or simply not?

@vovac12 Do you think there is a use case that we would need sudo access?

wer1st: Ok

wer1st (approved):

PR #1144, opened 2024-07-24
refactor: migration to polkadot sdk 1.1.0

i3ima:
This is the tracking PR for the upgrade of polkadot-sdk. 
_At the moment_ I can recall of the following issues:
- [ ] Some tests and benchmarks don't pass
- [ ] Local paths are used in patch sections of _workspace_-level `Cargo.toml` (i.e you are forced to have cloned repos of dependencies one directory up)
- [ ] Bunch of warnings, some of which can't be fixed by simple `cargo fix` -- It breaks things (I suspect it's because of procedural macros _or_ other conditional items visibility problem)

Not so issues, but important notes:
- This probably doesn't break anything but should be mentioned anyways. One of the packages that _1.1.0_ uses is `bandersnatch_vrfs` (from [ring-vrf](https://github.com/w3f/ring-vrf) which contains various important cryptography functions that are used in blocks production. It specifies this dependency via git repository and commit hash (`3119f51`). At this specific commit hash it causes compiler error because of [broken](https://github.com/w3f/ring-vrf/blob/3119f51b54b69308abfb0671f6176cb125ae1bf1/bandersnatch_vrfs/Cargo.toml#L22) `Cargo.toml`. So I [decided to bump](https://github.com/sora-xor/polkadot-sdk/commit/64051b489511c70ec67faf964147b792eb38f7b6) it

The lists will be updated as we progress


i3ima:
Currently this PR is not _fully_ ready for review as it's not very clean. Following things will be done next:
- [ ] Cleanup rustc & clippy warnings
- [ ] Test on predev

PR #1153, opened 2024-07-31
Update step quote verification

wer1st:
Now it accepted for `DiscreteQuotation` to have some slippage between chunk prices, not the strong order.


wer1st:
Was moved into #1141

pallets/liquidity-proxy/src/liquidity_aggregator/selector.rs
```diff
@@ -69,10 +68,8 @@ where
         &mut self,
         source: LiquiditySourceType,
         discrete_quotation: DiscreteQuotation<AssetIdOf<T>, Balance>,
```
vovac12, line 70: Maybe call this method `add_source_unchecked`?

wer1st: For what?

vovac12: To show that this function does not perform any checks before adding liquidity and it's caller's responsobility to perform any checks

wer1st: It makes sense only when we have another "checked" function.

vovac12 (approved):

Alexey-N-Chernyshov (approved):

Issue #1162, opened 2024-08-07
[Tech]: Chameleon XOR: VAL and PSWAP


Tieumsan:
### Problem statement

To optimise liquidity management synergies with Kensetsu assets, we want to add support for the (K)XOR-VAL and (K)XOR-PSWAP chameleon pools.

### Definition of Done

- [ ] (K)XOR-VAL chameleon pool
- [ ] (K)XOR-PSWAP chameleon pool

PR #1167, opened 2024-08-12
Prepare release 3.5.2

vovac12:


ZlayaMorda (approved):

i3ima (approved):

runtime/src/lib.rs
```diff
@@ -1028,7 +1028,7 @@ parameter_type_with_key! {
 parameter_type_with_key! {
     pub GetChameleonPools: |base: AssetId| -> Option<(AssetId, sp_std::collections::btree_set::BTreeSet<AssetId>)> {
         if *base == common::XOR {
-            Some((common::KXOR, [common::ETH, common::VAL, common::PSWAP].into_iter().collect()))
```
wer1st, line 1031: Do you need to completely revert this change or just deliver it later?
If later, please split it by feature flag.
If completely revert, please add coment to #1162 & #1163

wer1st (changes requested):

runtime/src/lib.rs
vovac12: Fixed

wer1st (approved):

PR #1187, opened 2024-09-03
Prepare release 3.6.2

vovac12:


HaidarJbeily7 (approved):

runtime/src/tests/liquidity_proxy.rs
```diff
@@ -52,7 +52,9 @@ pub fn ensure_balances(account_id: AccountId, assets: Vec<(AssetId, Balance)>) {
     }
 }
 
+// TODO: remove if we will completely remove Chameleon pools functionality
```
wer1st, line 55: In according with comment it looks like you have to remove these tests

wer1st (commented):
Don't forget about benchmarks

runtime/src/lib.rs
wer1st, line 1: And what about spec version?

vovac12: Thanks

runtime/src/tests/liquidity_proxy.rs
wer1st: Ok. Probably we don't need these comments in this case

PR #1213, opened 2024-09-25
Update buy back and burn strategy

vovac12:

### DIFF
--- a/pallets/xor-fee/src/lib.rs
+++ b/pallets/xor-fee/src/lib.rs
@@ -242,7 +242,7 @@ where
                     referrer_portion.into(),
                 ));
             } else {
-                total_xor_to_vxor += referrer_xor.peek();
+                total_xor_to_vxor = total_xor_to_vxor.saturating_add(referrer_xor.peek());
             }
 
             // TODO: decide what should be done with XOR if there is no referrer.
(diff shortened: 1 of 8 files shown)


Alexey-N-Chernyshov (approved):

pallets/xor-fee/src/lib.rs
```diff
@@ -239,16 +241,27 @@ where
                     referrer,
                     referrer_portion.into(),
                 ));
+            } else {
+                total_xor_to_vxor += referrer_xor.peek();
```
wer1st, line 245: checked or saturating add

wer1st: The same for all code below.
And above

runtime/src/tests/remote.rs
```diff
@@ -9,7 +9,7 @@ use std::env::var;
 async fn run_migrations() {
     sp_tracing::try_init_simple();
     let transport: Transport = var("WS")
-        .unwrap_or("wss://ws.framenode-0.a1.sora2.soramitsu.co.jp:443".to_string())
+        .unwrap_or("wss://ws.framenode-2.s1.stg1.sora2.soramitsu.co.jp:443".to_string())
```
wer1st, line 12: Why?

wer1st: Btw, it is better to introduce special consts for such cases.

wer1st (approved):

PR #1223, opened 2024-10-01
Fix migration

wer1st:


runtime/src/migrations.rs
```diff
@@ -28,4 +28,4 @@
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-pub type Migrations = (price_tools::migrations::v3::AddFastPriceInfos<crate::Runtime>,);
+pub type Migrations = ();
```
wer1st, line 31: Migration on Stage & Prod was done in the previous release. It is not necessary to call it again.

PR #1229, opened 2024-10-04
1198 automatic vesting support

ZlayaMorda:
* add storages for auto claim

* add hook for handling auto claim

* add tests

* max count of claims per hook based on weight of auto claim and max weight value


N1ghtStorm (approved):

runtime/src/tests/xor_fee.rs
```diff
@@ -28,6 +28,7 @@
 // STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+#![allow(unused_imports)]
```
Alexey-N-Chernyshov, line 31: It is better to use `feature` flag for import used in `wip`.

Alexey-N-Chernyshov (approved):
Please, be careful with `allow`.

PR #1237, opened 2024-10-14
Fix pending schedule params

ZlayaMorda:
* No need to change start after set up, so removed it

* No need to use None manager, so removed it

* Tests and benchmarks been changed according to new api

### DIFF
--- a/pallets/vested-rewards/src/lib.rs
+++ b/pallets/vested-rewards/src/lib.rs
@@ -292,14 +292,13 @@ impl<T: Config> Pallet<T> {
         // Independent logic for some schedules, so implementation only there
         match filter_schedule {
             VestingScheduleOf::<T>::LinearPendingVestingSchedule(filter_schedule) => {
-                filter_schedule.manager_id = Some(manager);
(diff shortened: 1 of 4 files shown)


pallets/vested-rewards/src/lib.rs
```diff
@@ -286,20 +286,23 @@ impl<T: Config> Pallet<T> {
...
         // Independent logic for some schedules, so implementation only there
         match filter_schedule {
             VestingScheduleOf::<T>::LinearPendingVestingSchedule(filter_schedule) => {
-                filter_schedule.manager_id = Some(manager);
+                filter_schedule.manager_id = manager;
+                let mut start = <frame_system::Pallet<T>>::block_number();
+                if let Some(input_start) = filter_schedule.start {
+                    start = input_start;
+                    filter_schedule.start = None;
+                }
```
wer1st, line 299:
```suggestion
                if let Some(input_start) = filter_schedule.start.take() {
                    start = input_start;
                }
```

wer1st (approved):

N1ghtStorm (approved):

PR #1239, opened 2024-10-15
Fix after `cargo update`

shlavik:
`wasm-bindgen` version fixed + new `Cargo.lock`


wer1st (approved):

.github/workflows/static_analysis.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
       # v2.20.16
```
ZlayaMorda, line 30:
```suggestion
      # latest v2
```

ZlayaMorda (approved):

PR #1247, opened 2024-10-22
Enable TON bridge on stage

vovac12:


runtime/Cargo.toml
```diff
@@ -310,16 +310,18 @@ wip = [
...
-    "bridge-channel",
     "pallet-beefy",
     "pallet-beefy-mmr",
     "xor-fee/wip",
-    "soratopia",
     "vested-rewards/wip",
 ]
 
-stage = ["framenode-chain-spec/stage"]
+stage = [
```
ZlayaMorda, line 319: looks like only `framenode-chain-spec` has the feature

ZlayaMorda (approved):

runtime/src/lib.rs
```diff
@@ -257,10 +255,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("sora-substrate"),
     impl_name: create_runtime_str!("sora-substrate"),
     authoring_version: 1,
-    spec_version: 100,
+    spec_version: 101,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
-    transaction_version: 100,
+    transaction_version: 101,
```
wer1st, line 261: Do you want  to deploy it on Stage without new version? Like 4.2.1

vovac12: Yes, because for now we just want to release it on stage and time for release on mainnet is not defined yet

runtime/Cargo.toml
vovac12: Yes, because other pallets is optional and it's enough to just enable them

wer1st (approved):

PR #1262, opened 2024-11-05
Prepare release 4.3.2

vovac12:
### Notes
* Band rate stale period was updated from 5 minutes to 10 minutes. Stale block period is not changed, so migration is not required
* Migration will be repeated

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -255,10 +255,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     spec_name: create_runtime_str!("sora-substrate"),
     impl_name: create_runtime_str!("sora-substrate"),
     authoring_version: 1,
-    spec_version: 103,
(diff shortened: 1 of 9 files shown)


runtime/src/lib.rs
```diff
@@ -1969,7 +1969,7 @@ impl oracle_proxy::Config for Runtime {
 }
 
 parameter_types! {
-    pub const GetBandRateStalePeriod: Moment = 60*5*1000; // 5 minutes
+    pub const GetBandRateStalePeriod: Moment = 60*10*1000; // 10 minutes
```
shlavik, line 1972: should this also be replaced in mocks for consistency?

shlavik (approved):

runtime/src/lib.rs
vovac12: Done

PR #1282, opened 2024-11-27
Presto initialize

wer1st:
Presto pallet initialization.

Developed:
- Add / remove managers
- Add / remove auditors
- Mint / burn / send PRUSD
- Requests workflow


pallets/presto/src/requests.rs
```diff
@@ -0,0 +1,260 @@
...
+        self.owner == other.owner
+            && self.time == other.time
+            && self.amount == other.amount
+            && self.payment_reference == other.payment_reference
+            && self.details == other.details
+            && self.status == other.status
+    }
+}
+
+impl<T: Config> Eq for DepositRequest<T> {}
```
wer1st, line 134: It is necessary to implement these traits manually, because there are some problems with `derive` for `T::MaxRequestPaymentReferenceSize` & `T::MaxRequestDetailsSize`

common/src/primitives.rs
```diff
@@ -200,6 +200,8 @@ mod _allowed_deprecated {
         KGOLD = 13,
         KXOR = 14,
         KARMA = 15,
+        #[cfg(feature = "wip")] // presto
```
vovac12, line 203: I think it's better to not use feature flag here, because it will make troubles with type generation on frontend side. We can just "reserve" this asset here without registration and when we're ready we will register it

pallets/presto/src/requests.rs
vovac12: You can also use derivative crate to solve this. [Here](https://github.com/sora-xor/sora2-common/blob/f38ef9d584d6dfe341af5b2788c5446528a53bf4/pallets/types/src/types.rs#L333) is example

common/src/primitives.rs
wer1st: There is a chance that the feature will be declined by WB

shlavik (approved):

vovac12 (approved):

PR #1297, opened 2024-12-10
Fix: cover new fee events with wip

ZlayaMorda:
Cover with wip to solve the SubQuery error

### DIFF
--- a/pallets/xor-fee/src/lib.rs
+++ b/pallets/xor-fee/src/lib.rs
@@ -341,11 +341,8 @@ where
                         TransactionValidityError::Invalid(InvalidTransaction::Payment)
                     })?;
 
-                    let xor_id = T::XorId::get();
-
                     Self::deposit_event(Event::FeeWithdrawn(
                         fee_source,
-                        xor_id,
                         adjusted_paid.peek().into(),
                     ));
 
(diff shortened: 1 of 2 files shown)


pallets/xor-fee/src/lib.rs
```diff
@@ -1264,10 +1268,18 @@ pub mod pallet {
...
         FeeWithdrawn(AccountIdOf<T>, AssetIdOf<T>, Balance),
+        #[cfg(feature = "wip")] // Xorless fee
         /// The portion of fee is sent to the referrer. [Referral, Referrer, AssetId, Amount]
         ReferrerRewarded(AccountIdOf<T>, AccountIdOf<T>, AssetIdOf<T>, Balance),
+        #[cfg(not(feature = "wip"))] // Xorless fee
+        /// Fee has been withdrawn from user. [Account Id to withdraw from, Fee Amount]
+        FeeWithdrawn(AccountIdOf<T>, Balance),
+        #[cfg(not(feature = "wip"))] // Xorless fee
+        /// The portion of fee is sent to the referrer. [Referral, Referrer, Amount]
+        ReferrerRewarded(AccountIdOf<T>, AccountIdOf<T>, Balance),
```
wer1st, line 1282: If we change the structure of event - it could broke indexers.
Probably it is better to add new events with new structure and just keep deprecated events for backward compatibility

ZlayaMorda: Yeah, that's why cover with wip.
Going update indexer. We don't want add separate events.
And asset id even now used in indexer, but only for xor.

vovac12 (approved):

shlavik (approved):

PR #1302, opened 2024-12-12
Feature/apollo update collateral factor

DanijelCBS:
Add collateral factor on Apollo for highly volatile and risky assets like KUSD.


pallets/apollo-platform/src/lib.rs
```diff
@@ -1565,6 +1642,33 @@ pub mod pallet {
...
+                    .unwrap_or(0),
+            )
+        }
+
+        /// Calculate the total existing collateral for a specific user and asset
+        fn calculate_total_existing_collateral(
+            user: &AccountIdOf<T>,
+            collateral_asset: &AssetIdOf<T>,
+        ) -> Balance {
+            <UserBorrowingInfo<T>>::iter()
```
vovac12, line 1661: I think it's better to aggregate total collateral amount in separate storage

CvijanCBS: Hello,
Thanks for the suggestion, we will do it, but it will take some time since we need to write migrations as well, to migrate the state of current users to the new storage that we create for it.

CvijanCBS (approved):

pallets/apollo-platform/src/lib.rs
```diff
@@ -1325,6 +1455,22 @@ pub mod pallet {
             )
         }
 
+        fn on_runtime_upgrade() -> Weight {
+            if Self::pallet_storage_version() == StorageVersion::V1 {
+                sp_runtime::runtime_logger::RuntimeLogger::init();
```
vovac12, line 1460: Runtime logging is enabled by default, don't need to enable it here

pallets/apollo-platform/src/lib.rs
```diff
@@ -1325,6 +1455,22 @@ pub mod pallet {
             )
         }
 
+        fn on_runtime_upgrade() -> Weight {
```
vovac12, line 1458: Please move runtime upgrade to empty struct with `OnRuntimeUpgrade` implementation and add it to runtime/src/migrations.rs. [Example](https://github.com/sora-xor/sora2-network/blob/master/pallets/multicollateral-bonding-curve-pool/src/migrations/v4.rs)

vovac12 (changes requested):

pallets/apollo-platform/src/lib.rs
```diff
@@ -120,22 +110,15 @@ pub mod pallet {
 
     pub type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
 
+    /// The current storage version.
+    pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(0);
```
vovac12, line 114: Here should be current version
```suggestion
    pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
```

pallets/apollo-platform/src/migrations.rs
```diff
@@ -1,44 +1,89 @@
...
+            "Migrated {} user total collateral entries",
+            total_migrated_entries
+        );
+
+        // Update storage version
+        StorageVersion::new(1).put::<Pallet<T>>();
+
+        // Calculate and return weight
+        <T as frame_system::Config>::DbWeight::get().reads_writes(
+            total_migrated_entries as u64 * 2, // read old and new storage
```
vovac12, line 61: Total entries calculation also reads all storage items
```suggestion
            total_migrated_entries as u64 * 3, // read old and new storage
```

vovac12 (approved):

vovac12 (approved):

PR #1304, opened 2024-12-18
Change fee to small fee for vesting

ZlayaMorda:
each claim counts as SMALL_FEE
vested_transfer default fee is SMALL_FEE
logic for vested_transfer fee calculation is the same

It is necessary to reduce the cost

### DIFF
--- a/runtime/src/xor_fee_impls.rs
+++ b/runtime/src/xor_fee_impls.rs
@@ -190,7 +190,8 @@ impl CustomFees {
             | RuntimeCall::Council(
                 pallet_collective::Call::close { .. } | pallet_collective::Call::propose { .. },
             )
-            | RuntimeCall::VestedRewards(vested_rewards::Call::vested_transfer { .. }) => {
+            | RuntimeCall::VestedRewards(vested_rewards::Call::vested_transfer { .. })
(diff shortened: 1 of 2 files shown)


runtime/src/xor_fee_impls.rs
```diff
@@ -234,16 +235,10 @@ impl xor_fee::ApplyCustomFees<RuntimeCall, AccountId> for CustomFees {
             RuntimeCall::VestedRewards(vested_rewards::Call::vested_transfer {
                 schedule, ..
             }) => {
-                let claim_fee = pallet_transaction_payment::Pallet::<Runtime>::weight_to_fee(
-                    <Runtime as Config>::WeightInfo::claim_unlocked(),
-                );
+                let claim_fee = fee;
```
shlavik, line 237: claim_fee variable looks redundant

shlavik (approved):

runtime/src/xor_fee_impls.rs
ZlayaMorda: Yeah, add comment with claim fee

vovac12 (approved):

vovac12 (approved):

PR #1336, opened 2025-07-29
refactor: reduce weight of `staking.payoutStakers` to 1/10

### LINKED ISSUE #1335: [Tech]: High Payout Transaction Fees
### Problem statement

-Validators doesnt run payout transactions
-Validators doesnt make money nor cover their server costs

### Description

-Due to high cost of payout transactions, validators are force to making these transactions
-Making payout transactions creates extra maintenance or needing for creating scripts
-Due to low transaction count val rewards are going down in usd value. This causes validators to not wanting to run payout transactions which results in with people not getting val rewards. 
-This payout transaction fees are actually comes back as val but since people dont run these transactions, it doesnt so decreasing this value wont change anything about val rewards at all. Right now only 17/69 validator made(or made by stakers) payment in last 48 hours
-Due to not profiting from running validator, they might shutdown their validator nodes

### Definition of Done

Either of these
-Payment transaction fee value should be reduced by 100 times the current value 
-Payout should be automated by the system

### Requirements

Changing the value is just changing the weight. But making system automaticly pay might be more complicated

0x009922:
Closes #1335

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -659,7 +659,64 @@ impl pallet_staking::Config for Runtime {
     type OnStakerSlash = ();
     type HistoryDepth = frame_support::traits::ConstU32<84>;
     type TargetList = pallet_staking::UseValidatorsMap<Self>;
-    type WeightInfo = ();
+    type WeightInfo = staking_weight::OverrideStackingWeightInfo;
+}
+
+pub mod staking_weight {
+    use super::*;
+    use pallet_staking::WeightInfo;
+
+    pub struct OverrideStackingWeightInfo;
+
+    macro_rules! delegate {
+        ($id:ident) => {
+            fn $id() -> Weight { <() as WeightInfo>::$id() }
+        };
+        ($fun:ident($($arg:ident: $ty:ty),+)) => {
+            fn $fun($($arg: $ty),+) -> Weight { <() as WeightInfo>::$fun($($arg),+) }
+        };
+    }
+
+    impl WeightInfo for OverrideStackingWeightInfo {
+        fn payout_stakers() -> Weight {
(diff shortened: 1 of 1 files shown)


0x009922:
I am not sure whether this approach would solve the problem.

I made some experiments locally with the test net, executing `staking.payoutStakers` pallet:

| weight change | fee change                              | fee           | result weight |
| ------------- | --------------------------------------- | ------------- | ------------- |
| _unchanged_   | _unchanged_, `0.0007` XOR (`SMALL_FEE`) | `420.000` XOR | `199,840,000` |
| `x / 10`      | _unchanged_                             | `420.000` XOR | `199,840,000` |
| `x / 1000`    | _unchanged_                             | `420.000` XOR | `99,940,000`  |
| `x * 999`     | _unchanged_                             | `420.000` XOR | `803,873,219` |
| `x * 999`     | `0.00007` XOR (`MINIMAL_FEE`)           | `42.0000` XOR **(!)** | `803,873,219` |

What you can see here:

- Simply multiplying base weight does not change the resulting weight with the same ratio. Dividing by 10 did not change anything. 
	- Dividing by 1000 decreased the result by 2 (199k to 99k)
	- Multiplying by 999 increased the result only by ~4 (199k to 803k)
- Somehow, weight does not affect the fee. This may be due to `runtime/src/xor_fee_impls.rs`, which assigns the fee for `payoutStakers` is to a **constant** `SMALL_FEE`.
- Applied fee only changed when I changed this assigned constant fee.

Also, I noticed that there is a huge fee (in production Sora) on plain transfers (`assets.transfer`) - 12 Tril XOR, same as for `staking.payoutStakers`. This makes me wonder if only the `staking.payoutStakers` extrinsic must be fixed, or there should be some global fix?

PR #1337, opened 2025-07-30
refactor: reduce `staking.payoutStakers` fee to 1/10th

### LINKED ISSUE #1335: [Tech]: High Payout Transaction Fees

0x009922:
Closes #1335

Reducing the fee as per https://github.com/sora-xor/sora2-network/issues/1335#issuecomment-3004262480

The other PR, #1336, does not solve the problem (as said in https://github.com/sora-xor/sora2-network/pull/1336#issuecomment-3131039636).

This PR simply switches the fee for `staking.payoutStakers` from `SMALL_FEE` constant to `MINIMAL_FEE`, which is 10 times smaller.

### DIFF
--- a/runtime/src/xor_fee_impls.rs
+++ b/runtime/src/xor_fee_impls.rs
@@ -220,7 +220,6 @@ impl CustomFees {
             | RuntimeCall::MulticollateralBondingCurvePool(..)
             | RuntimeCall::PoolXYK(..)
             | RuntimeCall::Rewards(..)
-            | RuntimeCall::Staking(pallet_staking::Call::payout_stakers { .. })
             | RuntimeCall::TradingPair(..)
             | RuntimeCall::Referrals(..)
             | RuntimeCall::OrderBook(..)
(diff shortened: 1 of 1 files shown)


ra9mls (approved):

WRRicht3r (approved):

stefashkaa (approved):

0x009922:
Superseded by #1340

PR #1338, opened 2025-07-30
refactor: set on-chain election fallback

0x009922:
This PR sets `Fallback` for `pallet_election_provider_multi_phase` to on-chain election (same as for `GovernanceFallback`). As per @vovac12, this is supposed to solve the eras issue (which are not switching automatically due to failing election fallback).


0x009922:
Can confirm that election no longer falls into emergency.

**Without the fix:** `Off -> Signed -> Unsigned: [true, 5] -> Emergency`

**With it:** `Off -> Signed -> Unsigned: [true, 5] -> Off -> Signed -> Unsigned: [true, 25] -> Off -> ...`

ra9mls (approved):

WRRicht3r (approved):

